PR #34, opened 2022-04-25
Add skeleton react app

jadmsaadaot:


met-web/src/App.js
```diff
@@ -0,0 +1,45 @@
+import "./App.css";
+import React, { useEffect } from "react";
+import Header from "./components/layout/Header";
+import { BrowserRouter as Router } from "react-router-dom";
+import { useDispatch, useSelector } from "react-redux";
+import BaseRouting from "./components/Routing/BaseRouting";
+import { CircularProgress, Grid } from "@mui/material";
+import UserService from "./services/UserServices";
+
```
djnunez-aot, line 9: Looks good!

djnunez-aot (approved):

tiagograf-AOT:
I have a few suggestions that we can discuss but I think you can drop the "met-" in the folder name and keep it as just "web" since met is already implicit in the repository name.

As for the code itself, I personally like Typescript development over Javascript, in my opinion it helps with maintainability, 
@nikhila-aot what do you think?

PR #53, opened 2022-04-29
Add docker files to apps

tiagograf-AOT:
Adds requirements files to met-api
Fixes folder structure in met-api


met-web/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# set working directory
+WORKDIR /app
+
+# add `/app/node_modules/.bin` to $PATH
+ENV PATH /app/node_modules/.bin:$PATH
+
+# install app dependencies
+COPY package.json ./
+COPY package-lock.json ./
```
nikhila-aot, line 12: is the package-lock needed ?

djnunez-aot (approved):

met-web/Dockerfile
tiagograf-AOT: its optional but it makes sure that the packages are being downloaded from the same source as when it was added

nikhila-aot (approved):

PR #94, opened 2022-05-09
Add Create engagement endpoint

jadmsaadaot:


met-api/app/services/engagement_service.py
```diff
@@ -0,0 +1,30 @@
+from app.models.engagement import Engagement
+
+class engagement_service:
+    """ Engagement management service
+    """
+    otherdateformat = '%Y-%m-%d'
+
+    def get_engagement(self, engagement_id):
+        request_engagement = Engagement().get_engagement(engagement_id)
```
tiagograf-AOT, line 9: I'm not familiar with python but does Engagement() means its creating a new instance of the engagement class?

tiagograf-AOT (approved):
It looks good but I think we should have a discussion about dependency injection / factory design patterns

met-api/app/services/engagement_service.py
jadmsaadaot: Good eye, I have changed it to Engagement without calling it. Bit of research is perhaps needed here to make sure whether  python creates an Engagement instance at import and not interprets Engagement.get_engagement same as Engagement().get_engagement

PR #96, opened 2022-05-10
Landing Page Draft 1

djnunez-aot:
Added met header, engagement table, responsiveness


met-web/src/components/layout/Header.tsx
```diff
@@ -1,36 +1,119 @@
...
+    align-items: flex-end;
+    justify-content: flex-start;
+  `;
+
+  const LogoutContainer = styled.div`
+    display: flex;
+    flex: 1;
+    align-items: center;
+    justify-content: center;
+  `;
```
jadmsaadaot, line 51: it seems that every element in the app bar has a container with a different flex. What is the reason they don't exist in the same flex container and aligned together under a single rule.

met-web/src/components/layout/Header.tsx
```diff
@@ -1,36 +1,119 @@
...
+
```
jadmsaadaot, line 52: It seems that every component in the header has its own flex container and its own flex alignment. Are we not able to have them under the same alignment rules in one flex container? Also as of now they are a bit misaligned

met-web/src/components/layout/Header.tsx
```diff
@@ -1,36 +1,119 @@
...
+            noWrap
+            component="div"
+            sx={{ mr: 2 }}
+          >
             MET
           </Typography>
         </Toolbar>
-      </Container>
+      </TitleContainer>
+      <LogoutContainer>
```
jadmsaadaot, line 97: is this component part of the design?

met-web/src/components/layout/Table.tsx
```diff
@@ -0,0 +1,436 @@
+import * as React from "react";
```
jadmsaadaot, line 1: Noticed several unused imports, unused methods, commented out blocks of code. Also it would be good if all our functions remained arrow functions for consistency. Component seems to need some cleanup from all the extra features in it that are not used as well.

met-web/src/constants/constants.ts
```diff
@@ -1,9 +1,13 @@
+
+
 const getFromEnv = (key: string, defaultValue: string = "") => {
   if (!key) return "";
 
   return process.env[key] || defaultValue;
 };
 
+console.log(process.env)
```
jadmsaadaot, line 9: do we need this?

met-web/src/constants/constants.ts
```diff
@@ -25,3 +31,5 @@ export const KEYCLOAK_AUTH_URL = `${KEYCLOAK_URL}`;
 
 export const ANONYMOUS_ID = getFromEnv("REACT_APP_ANONYMOUS_ID");
 export const ANONYMOUS_USER = "anonymous";
+
```
jadmsaadaot, line 34: unnecessary extra space

met-web/src/routes/LandingPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const DrawerContainer = styled.div`
+  display: flex;
+  flex: 0.1;
+  height: 90vh;
+  align-items: center;
+  justify-content: center;
+`;
+
+const TableContainer = styled.div`
```
jadmsaadaot, line 21: We can consider Material UI <Grid> here to avoid extra css and multiple flex containers

met-web/src/services/UserServices.ts
```diff
@@ -36,24 +38,24 @@ const initKeycloak = (dispatch: Dispatch<any>) => {
...
+
         dispatch(userToken(KeycloakData.token));
 
         KeycloakData.loadUserInfo().then((res: UserDetail) => {
           dispatch(userDetails(res));
           dispatch(userAuthorization(true));
         });
-        dispatch(userAuthentication(authenticated));
+
+        dispatch(userAuthentication(KeycloakData.authenticated ? true : false));
```
jadmsaadaot, line 58: how is KeycloakData.authenticated different that the authenticated prop passed to this method in the beginning?

met-web/src/components/Login/index.tsx
```diff
@@ -1,9 +1,16 @@
...
 import UserService from "../../services/UserServices";
 import { Grid } from "@mui/material";
 import Button from "@mui/material/Button";
+import { useAppSelector, useAppDispatch } from "../../hooks";
 
 const Login = () => {
+  const [isLoggedIn, setIsLoggedIn] = useState<boolean>(false);
+
+  useEffect(() => {
+    setIsLoggedIn(UserService.isLoggedIn);
```
jadmsaadaot, line 11: useEffect always needs seconds parameter. Also user shouldn't be on Login page if user is not logged in. It seems out of place to check (that user is logged in) in this component.

jadmsaadaot (dismissed):
It seems that   

const [isLoggedIn, setIsLoggedIn] = useState<boolean>(false);

  useEffect(() => {
    setIsLoggedIn(UserService.isLoggedIn)
})

has been used repeatedly. If the authentication in the Redux store is not working properly we should aim to fix that rather than using the above patch solution.

I also added other comments

tiagograf-AOT (approved):
I have a few suggestions regarding the file names/structure, also a few nits throughout the code but I'm good to have this PR merged since its already way too big, the first thing we can work on is some sort of Linting rules to make the code format standard, we should have a chat Monday about it

Good job on figuring out the issues with redux!

saravanpa-aot:
This changes have been incorporated in another PR.so closing it

PR #148, opened 2022-05-25
added api and service layer tests

saravanpa-aot:
Added tests for service and api layer
removed unnecessary code


met-api/src/met_api/__init__.py
```diff
@@ -37,7 +37,7 @@
     xfo=xfo_value
 )
 
-@app.after_request
+#@app.after_request
 def set_secure_headers(response):
     """Set CORS headers for security."""
     secure_headers.framework.flask(response)
```
saravanpa-aot, line 43: Will come back to this.need more time to see how it works and it wasnt working in my local...Hope thats alrite

Issue #103, opened 2022-05-12
An Admin can edit a draft engagement


tiagograf-AOT:
As an Admin I would like to edit a draft engagement so that I can update an incorrect information like the dates or it's description.

- [ ] create end point
- [ ] create a link to engagements
- [ ] create and update engagement form

PR #165, opened 2022-05-30
Add loader and error callback

jadmsaadaot:
Issue #103 

Description of changes:
- Add loader for when saved engagement is being fetched
- Add error callback for fetch engagement call


met-api/src/met_api/models/user.py
```diff
@@ -24,14 +24,14 @@ class User(db.Model):
     engagement_id = db.relationship('Engagement', backref='user', cascade="all, delete")
 
     @classmethod
-    def find_by_id(cls, id):
+    def find_by_id(cls, _id):
```
saravanpa-aot, line 27: any reason for making it _id ? prefixed underscores means its kind of private variable

saravanpa-aot (approved):

PR #186, opened 2022-06-02
Alter migration scripts to not specify description length

jadmsaadaot:


djnunez-aot (approved):

tiagograf-AOT:
Im not sure how our migrations work, are they cumulative or do they reshape the Sb every time? If it’s cumulative removing those statements from past migration scripts won’t automatically fix the places where those migrations were already run @jadmsaadaot @saravanpa-aot can you guys confirm this?

saravanpa-aot:
> Im not sure how our migrations work, are they cumulative or do they reshape the Sb every time? If it’s cumulative removing those statements from past migration scripts won’t automatically fix the places where those migrations were already run @jadmsaadaot @saravanpa-aot can you guys confirm this?

Only the new ones run.. so the if we modify the old ones wont be of any impact..
But in our case , since its not run in TEST/Prod , every script is new..so whatever you suggested will work..overhead is that in DEV , we will have to manually run

jadmsaadaot:
> > Im not sure how our migrations work, are they cumulative or do they reshape the Sb every time? If it’s cumulative removing those statements from past migration scripts won’t automatically fix the places where those migrations were already run @jadmsaadaot @saravanpa-aot can you guys confirm this?
> 
> Only the new ones run.. so the if we modify the old ones wont be of any impact.. But in our case , since its not run in TEST/Prod , every script is new..so whatever you suggested will work..overhead is that in DEV , we will have to manually run

@tiagograf-AOT @saravanpa-aot , yes this is what I thought. If we already have a prod deployment I would've created a new migration. But since we're still at initiation phase, I just thought it's easier to update the current scripts. That it might be cleaner to define it from the start as unlimited rather than limited to 50 then update that field to fix it.

tiagograf-AOT (approved):

Issue #173, opened 2022-05-31
Update MUI Theme not to Uppercase all button text


jadmsaadaot:
See if we can add an option in the MUI Theme to stop Buttons from having uppercase automatically. If that cannot be done, we might need to use Styled to customize MuiButton to not have uppercase, check what we did for Box and styled it to MetBox in the common components

Issue #168, opened 2022-05-31
Change UI project to use SCSS instead of CSS


saravanpa-aot:
AS of now , react project uses CSS ;but using a CSS processor like SCSS is much preferred 

^^ @jadmsaadaot  @djnunez-aot

PR #191, opened 2022-06-05
CSS to SCSS

djnunez-aot:
Changing a few files and temporary fix to the material ui & formio style conflict


met-web/src/App.scss
```diff
@@ -0,0 +1,6 @@
+@charset "UTF-8";
```
jadmsaadaot, line 1: Unrelated to this task, should be part of #173

met-web/src/App.tsx
```diff
@@ -49,17 +49,19 @@ const App = () => {
 
     return (
         <Router>
-            <Box sx={{ display: 'flex' }}>
-                <LoggedInHeader drawerWidth={drawerWidth} />
+            <StyledEngineProvider injectFirst>
```
jadmsaadaot, line 52: Perhaps better to be put at the index.js file in the root folder and not here

met-web/src/components/Form/Form.scss
```diff
@@ -0,0 +1,3 @@
+.FormIO{
+    @use "FormStyles.scss";
```
jadmsaadaot, line 2: not related to this task

met-web/src/components/common/index.tsx
```diff
@@ -1,13 +1,13 @@
...
+import { Button as MuiButton, Container, Grid, CircularProgress } from '@mui/material';
 import styled from '@emotion/styled';
 
 export const RoundedButton = styled(MuiButton)(() => ({
     borderRadius: '23px',
 }));
 
-export const MetBox = styled(MuiPaper)(() => ({
-    border: `1px solid #606060`,
+export const MetBox = styled(Container)(() => ({
```
jadmsaadaot, line 9: Why?
- Container adds padding left and right, while Paper does not. MetBox  was not used (in engagement form for example) with the intention of it behaving like a MUI Container.
- We would do better to avoid overwriting components in the common folder and create new ones if we needed them.

met-web/src/components/Form/View.tsx
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable */
 import React, { useEffect } from 'react';
+import './Form.scss';
```
jadmsaadaot, line 3: Out of scope for task #168

met-web/src/components/common/index.tsx
```diff
@@ -1,13 +1,13 @@
...
+    border: `1.5px solid #606060`,
```
jadmsaadaot, line 10: why change from 1px to 1.5px? 1px is as per the design https://xd.adobe.com/view/7360b654-4a25-4140-9670-a56f4adabe7a-2633/screen/32caf4d5-99c5-4df2-9e43-620e6b838203/specs/

met-web/src/index.scss
```diff
@@ -22,7 +22,7 @@
 
 body {
   margin: 0;
-  font-family: "BCSans-Regular";
```
jadmsaadaot, line 25: If we are not going to apply it, better remove it right?

jadmsaadaot (changes requested):
The transition looks good. I noticed some issues:
- Other tasks are done here and they should be separate with their own PR (FormIO and MUI Button)
- Risky change in the common components folder, should be better avoided. How does the Engagement Form in the create engagement page look like now?
- Provider could be better placed in another file
- 3 failing tests noticed

met-web/src/App.scss
djnunez-aot: Ok will make a separate PR

jadmsaadaot (approved):

tiagograf-AOT (approved):

Issue #216, opened 2022-06-09
Fix Status Column in Engagement List


tiagograf-AOT:
Expected:
Engagement status column shows "Draft"
Actual:
Engagement status column is always blank

PR #218, opened 2022-06-13
Engagement Status bug fix

djnunez-aot:
*Issue :#216*


met-web/src/components/engagement/EngagementForm.tsx
```diff
@@ -32,7 +32,7 @@ const EngagementForm = () => {
             fromDate: formatDate(savedEngagement.start_date),
             toDate: formatDate(savedEngagement.end_date),
             description: savedEngagement?.description || '',
-            status_id: savedEngagement?.status_id || '',
```
tiagograf-AOT, line 35: We need to find a way to load the "Draft' status from the engagement_status table, the "status_id" is not what we want to display since its a numeric field (1,2, etc..), using SQLAlchemy there must be a way to configure it so it would join tables as needed

met-web/src/components/engagement/ActionContext.tsx
```diff
@@ -41,7 +41,7 @@ export const ActionProvider = ({ children }: { children: JSX.Element }) => {
         name: '',
         description: '',
         rich_description: '',
-        status_id: '',
```
tiagograf-AOT, line 44: status_id should be numeric

tiagograf-AOT (changes requested):
We need some more changes to get this to work properly, basically when returning the engagement json I expect we have something like this { ..., name: 'Test Engagement', status: {status_id: 1, status: 'Draft'} }

PR #223, opened 2022-06-16
Update http API Service

jadmsaadaot:
Issue #119 
Description of changes:

- Create wrapper methods for CRUD operations in React which adds the jwt and the necessary default headers


met-web/src/apiManager/endpoints/index.ts
```diff
@@ -0,0 +1,9 @@
+import { AppConfig } from '../../config';
+
+const API = {
+    MET_GET_ENGAGEMENTS: `${AppConfig.apiUrl}/engagement/`,
+    MET_GET_ENGAGEMENT: `${AppConfig.apiUrl}/engagement/<engagement_id>`,
+    MET_CREATE_ENGAGEMENT: `${AppConfig.apiUrl}/engagement/`,
+    MET_UPDATE_ENGAGEMENT: `${AppConfig.apiUrl}/engagement/`,
+};
```
jadmsaadaot, line 8: Should the above GET, CREATE and Update engagements be made into one URL. I feared confusion and it maybe being harder to change in the future if we wanted a different URL (for some reason) for them

saravanpa-aot, line 8: Thanks for doing this..Much needed change..
But I think its safe to go with one URL ..99% we will have one rest end point at backend... so MET_ENGAGEMENTS will be good enough to support all rest verbs on that url..
Also do we need MET_? MET is kind of implicit..

met-web/src/apiManager/endpoints/index.ts
```diff
@@ -0,0 +1,9 @@
+import { AppConfig } from '../../config';
```
saravanpa-aot, line 1: Curious , is there a better way of importing than absolute imports?
Like Vue had @ which helps to use something like 
import AccountChangeMixin from '@/components/auth/mixins/AccountChangeMixin.vue'
I am not sure if react we can achieve this.but putting out this here so that you can check

met-web/src/services/engagementService/index.ts
```diff
@@ -22,16 +18,11 @@ export const getEngagement = async (
     errorCallback: (errorMessage: string) => void,
 ) => {
     try {
+        const url = replaceUrl(API.MET_GET_ENGAGEMENT, '<engagement_id>', String(engagementId));
```
saravanpa-aot, line 21: can https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals help here?

saravanpa-aot (approved):
Good Stuff..Thanks for the small PR ;easy to review
A few comments are in there ;which are improvements 
Otherwise it looks good...

met-web/src/apiManager/httpRequestHandler/index.ts
```diff
@@ -0,0 +1,30 @@
+import axios from 'axios';
+import UserService from '../../services/userService';
+
```
tiagograf-AOT, line 3: You could have a more typescript approach which is having a exporting a default class with multiple methods

tiagograf-AOT (dismissed):

met-web/src/services/engagementService/index.ts
jadmsaadaot: Here I'm actually defining the template of the URL as Base/engagement/<engagement_id> in the endpoints file. Here I'm just replacing the <engagement_id> bit. The template literals will just insert the variable from the beginning. `engagemnet/${id}`.

met-web/src/apiManager/httpRequestHandler/index.ts
jadmsaadaot: done

met-web/src/apiManager/endpoints/index.ts
jadmsaadaot: You mean letting go of the relative path with the '../../' ? I reviewed it on https://medium.com/geekculture/making-life-easier-with-absolute-imports-react-in-javascript-and-typescript-bbdab8a8a3a1. It seems doable and let's add a new ticket for it

met-web/src/apiManager/endpoints/index.ts
jadmsaadaot: I have a similar feeling but just feel it's difficult to name an endpoint and use for single and multiple operations. Like MET_ENGAGEMENTS or ENGAGEMENTS and use it for getting all engagements but also for creating one single engagement. What do you think?

tiagograf-AOT (approved):

saravanpa-aot (approved):

Issue #102, opened 2022-05-12
IDIR users must be saved DB


tiagograf-AOT:
After a admin user successfully login, it's information (name, id) must be saved in the met database so that it can be used for display.

- [ ] end point to save admin user data
- [ ] data base
- [ ] edit front end

sidtobias:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/102?planning-poker) with ZenHub @djnunez-aot @jadmsaadaot @tiagograf-AOT

PR #233, opened 2022-06-17
Save users to db

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/102

*Description of changes:*

Saves jwt user information to the database.


djnunez-aot (approved):

met-web/src/services/userService/index.ts
```diff
@@ -107,8 +110,18 @@ const hasRole = (role: string) => KeycloakData.hasResourceRole(role);
 
 const hasAdminRole = () => KeycloakData.hasResourceRole(AppConfig.keycloak.adminRole);
 
+const updateUser = async () => {
+    try {
+        await http.PutRequest(Endpoints.User.CREATE_UPDATE);
+        // TODO: update store with current user info.
+    } catch (e: unknown) {
+        console.error(e);
```
jadmsaadaot, line 118: should we consider logging the user out if this put request failed? Are there any big concerns here if this put request failed and the user was left to log in?

met-api/src/met_api/resources/user.py
```diff
@@ -0,0 +1,57 @@
...
+class UserController(Resource):
+    """User controller class."""
+
+    @staticmethod
+    # @TRACER.trace()
+    @cross_origin(origins=allowedorigins())
+    @auth.require
+    def put():
+        """Update or create a user."""
+        try:
```
saravanpa-aot, line 42: hm.. good approach..
one challenge will be if we ever need a user update from a staff ;say staff updating users first name , this might not work..
a put on {user_id} will work in both cases..but we have to make sure the authorisation

Issue #118, opened 2022-05-18
Add Notifications (Success and Error)


sidtobias:
As an Admin I would like to be notified about the completion of the actions that I take throughout the app (If I edit and save an engagement I expect to see a "Success" message)

PR #253, opened 2022-06-24
Setup Notifications 

djnunez-aot:
*Issue #118 :*
https://github.com/bcgov/met-public/issues/#118

Add notifications for create and update engagement using MUI Snackbar


jadmsaadaot (changes requested):
Notifications should be for all the application to use not just for the engagement component.

We can create a slice in redux, a notification slice, that has the open/closed status and the severity.

So whenever we want to have a notification we can dispatch(openNotificatio(<severity>)). And that action would mutate the notification status to open and will change the severity as specified in the action.

we can create another action in the slice closeNotification() which would change status to closed. We can dispatch that in the handleClose  function. And the Snackbar we can put it in the App.tsx file so it becomes at the whole application level

met-web/src/services/notificationService/notificationSlice.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const notificationSlice = createSlice({
+    name: 'notification',
+    initialState,
+    reducers: {
+        openNotification: (state: NotificationState, action: PayloadAction<NotificationState>) => {
+            state.open = action.payload.open;
+            state.severity = action.payload.severity;
+            state.text = action.payload.text;
+        },
+        closeNotification: (state: NotificationState) => {
```
jadmsaadaot, line 19: I would consider passing the initial state here so everything resets after it closes. (like text goes back to empty '')

jadmsaadaot: state = initialState maybe

jadmsaadaot (commented):
Looks great and clean. Just noticed one minor simple improvement potential when closing the notification.

met-web/src/services/notificationService/notificationSlice.tsx
djnunez-aot: We could do that but the update happens while the notification is on screen and it looks weird up to you

jadmsaadaot: Ok all good, this is fine then. I would suggest after merging this to do a bit of research on how we can make it multiple alerts. This is great, the drawback is that if we had multiple problems happening at the same time and multiple dispatches happening only one will show. We might not need that, I suppose most of the time there will be one notification showing anyway. Research on this, however, whether we can implement it or not would be helpful

jadmsaadaot (approved):

met-web/src/services/engagementService/engagementSlice.ts
```diff
@@ -10,7 +10,7 @@ export const engagementSlice = createSlice({
     name: 'engagement',
     initialState,
     reducers: {
-        setEngagements: (state, action: PayloadAction<Engagement[]>) => {
+        setEngagements: (state: EngagementState, action: PayloadAction<Engagement[]>) => {
```
tiagograf-AOT, line 13: I like the strongly typed stuff! thanks

tiagograf-AOT (approved):

Issue #257, opened 2022-06-28
Missing Engagement Preview Button


tiagograf-AOT:
Expected
After saving an engagement or when editing an existing engagement a preview button should be available. 

Actual
No preview button is available for the user.

PR #264, opened 2022-06-29
Add engagement preview button

tiagograf-AOT:
Refactor callbacks into promises

*Issue #:*
https://github.com/bcgov/met-public/issues/257

*Description of changes:*
Adds Preview button to engagement form
Refactor callbacks into promises

### DIFF
--- a/met-web/src/components/engagement/form/EngagementForm.tsx
+++ b/met-web/src/components/engagement/form/EngagementForm.tsx
@@ -115,34 +115,34 @@ const EngagementForm = () => {
         return Object.values(errors).some((isError: unknown) => isError);
     };
 
-    const handleSaveEngagement = (isPreview: boolean) => {
+    const handleSaveEngagement = async (isPreview: boolean) => {
         const hasErrors = validateForm();
 
         if (hasErrors) {
             return Promise.reject('Validation Errors');
         }
         if (isNewEngagement) {
-            handleCreateEngagementRequest({
+            const engagement = await handleCreateEngagementRequest({
                 ...engagementFormData,
                 richDescription: richDescription,
                 richContent: richContent,
-            }).then((engagement) => {
-                if (isPreview) {
-                    navigate(`/engagement/view/${engagement.id}`);
-                } else {
-                    navigate(`/engagement/form/${engagement.id}`);
-                }
             });
+
+            if (isPreview) {
+                navigate(`/engagement/view/${engagement.id}`);
+            } else {
+                navigate(`/engagement/form/${engagement.id}`);
+            }
         } else {
-            handleUpdateEngagementRequest({
+            const engagement = await handleUpdateEngagementRequest({
                 ...engagementFormData,
                 richDescription: richDescription,
                 richContent: richContent,
-            }).then((engagement) => {
-                if (isPreview) {
-                    navigate(`/engagement/view/${engagement.id}`);
-                }
             });
+
+            if (isPreview) {
+                navigate(`/engagement/view/${engagement.id}`);
+            }
         }
     };
 
--- a/met-web/src/components/engagement/form/EngagementForm.tsx
+++ b/met-web/src/components/engagement/form/EngagementForm.tsx
@@ -115,32 +115,47 @@ const EngagementForm = () => {
         return Object.values(errors).some((isError: unknown) => isError);
     };
 
-    const handleSaveEngagement = async (isPreview: boolean) => {
+    const handleCreateEngagement = async () => {
         const hasErrors = validateForm();
 
-        if (hasErrors) {
-            return Promise.reject('Validation Errors');
-        }
-        if (isNewEngagement) {
+        if (!hasErrors) {
             const engagement = await handleCreateEngagementRequest({
                 ...engagementFormData,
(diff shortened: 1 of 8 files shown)


met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -111,27 +114,35 @@ const EngagementForm = () => {
 
         return Object.values(errors).some((isError: unknown) => isError);
     };
-    const handleCreateEngagement = () => {
```
jadmsaadaot, line 114: I do see the value that it is more succinct to collapse create engagement and update engagement into one method instead of having repetition in two. However, I find it increases cognitive complexity inside of a method especially if we are using an if statement to determine its behavior.

I am hesitant to go that way following the suggested Functional Rules https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29. Namely "Do one thing"
and "fewer arguments".  That's why I separated them. It becomes easier to understand how each method behaves. There is the cost however to do two buttons with a conditional.

met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -111,27 +114,35 @@ const EngagementForm = () => {
...
-        const errorExists = validateForm();
 
-        if (!errorExists) {
-            handleCreateEngagementRequest({
+    const handleSaveEngagement = async (isPreview: boolean) => {
+        const hasErrors = validateForm();
+
+        if (hasErrors) {
+            return Promise.reject('Validation Errors');
```
jadmsaadaot, line 122: better than placing everything under this under an if statement if(!errorExists), maybe we can just return nothing here instead of rejecting a promise. Worried that might have to handle this promise rejection if we don't need it

met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -279,28 +290,19 @@ const EngagementForm = () => {
...
-                                >
-                                    Update Engagement Draft
-                                    {saving && <CircularProgress sx={{ marginLeft: 1 }} size={20} />}
-                                </Button>
-                            )}
+                        <Grid item xs={12}>
+                            <Button
+                                variant="contained"
+                                sx={{ marginRight: 1 }}
+                                onClick={() => handleSaveEngagement(false)}
```
jadmsaadaot, line 297: I think the design isn't to move to preview with the save button, it's to have another separate button to move to preview. Whether creating or updating I understood not to navigate away from the page since the user might need to do more changes.

jadmsaadaot (changes requested):
Detected increase of cognitive complexity

met-web/src/components/engagement/form/EngagementForm.tsx
tiagograf-AOT: Saving an engagement does not navigate away, hence the "Save & preview button"

met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -279,28 +290,19 @@ const EngagementForm = () => {
...
+                                disabled={isSaving}
+                            >
+                                {isNewEngagement ? 'Create Engagement Draft' : 'Update Engagement Draft'}
+                                {isSaving && <CircularProgress sx={{ marginLeft: 1 }} size={20} />}
+                            </Button>
+                            <Button variant="outlined" onClick={() => handleSaveEngagement(true)} disabled={isSaving}>
```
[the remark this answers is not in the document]
tiagograf-AOT: We need to save the engagement to view it currently, i don't understand your comment

met-web/src/components/engagement/form/EngagementForm.tsx
jadmsaadaot: My bad here, I misread navigate(`/engagement/form/${engagement.id}`);

met-web/src/components/engagement/form/EngagementForm.tsx
tiagograf-AOT: Changed

met-web/src/components/engagement/form/EngagementForm.tsx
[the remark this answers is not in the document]
tiagograf-AOT: not in the scope of this bug anyway

met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -111,27 +114,35 @@ const EngagementForm = () => {
...
             });
-        }
-    };
-
-    const handleUpdateEngagement = () => {
-        const errorExists = validateForm();
 
-        if (!errorExists) {
-            handleUpdateEngagementRequest({
+            if (isPreview) {
```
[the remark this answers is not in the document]
tiagograf-AOT: Changed

met-web/src/components/engagement/form/EngagementForm.tsx
tiagograf-AOT: updated

jadmsaadaot (approved):

saravanpa-aot (approved):

Issue #219, opened 2022-06-14
Create Survey/Forms API Endpoints


tiagograf-AOT:
- [ ] Create db table (Survey) with the following columns:
SurveyId UUID
Name varchar(50)
FormJSON JSON
created_date date
updated_date date
created_by UUID
updated_by UUID

- [ ] Create POST (Admin), GET (Anonymous), PUT (Admin), DELETE (Admin), GET ALL (Admin) endpoints

tiagograf-AOT:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/219?planning-poker) with ZenHub @jadmsaadaot @djnunez-aot @saravanpa-aot

PR #268, opened 2022-07-04
survey endpoints [In Progress]

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/219

*Description of changes:*
Adds new endpoints, schema and service for survey

### DIFF
--- a/met-api/src/met_api/resources/survey.py
+++ b/met-api/src/met_api/resources/survey.py
@@ -11,57 +11,57 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""API endpoints for managing an engagement resource."""
+"""API endpoints for managing an survey resource."""
 
 from flask import request
 from flask_cors import cross_origin
 from flask_restx import Namespace, Resource
 
 from met_api.auth import auth
-from met_api.schemas.engagement import EngagementSchema
-from met_api.services.engagement_service import EngagementService
+from met_api.schemas.survey import SurveySchema
+from met_api.services.survey_service import SurveyService
 from met_api.utils.action_result import ActionResult
 from met_api.utils.token_info import TokenInfo
 from met_api.utils.util import allowedorigins, cors_preflight
 
 
-API = Namespace('engagement', description='Endpoints for Engagements Management')
+API = Namespace('survey', description='Endpoints for Survey Management')
 """Custom exception messages
 """
 
 
 @cors_preflight('GET,OPTIONS')
-@API.route('/<engagement_id>')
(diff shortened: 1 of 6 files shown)


met-api/migrations/versions/fc570c0faace_survey_migration.py
```diff
@@ -0,0 +1,41 @@
...
+down_revision = 'ec504565fab3'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('survey',
+    sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
+    sa.Column('name', sa.String(length=50), nullable=True),
```
saravanpa-aot, line 23: we can throw in an index for name and id i think

met-api/src/met_api/models/survey.py
```diff
@@ -0,0 +1,71 @@
...
+        """Update engagement."""
+        update_fields = dict(
+            name=survey.get('name', None),
+            form_json=survey.get('form_json', None),
+            created_date=survey.get('created_date', None),
+            updated_date=survey.get('updated_date', None),
+            created_by=survey.get('created_by', None),
+            updated_by=survey.get('updated_by', None),
+        )
+        Survey.query.filter_by(id=survey.get('id', None)).update(update_fields)
```
saravanpa-aot, line 69: may be split into two lines and add a if check ..
incase if id is wrong and db doesnt have that survey

djnunez-aot (approved):

met-api/src/met_api/resources/survey.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+API = Namespace('survey', description='Endpoints for Survey Management')
+"""Custom exception messages
+"""
+
+
+@cors_preflight('GET,OPTIONS')
+@API.route('/<survey_id>')
+class GetSurvey(Resource):
```
saravanpa-aot, line 35: GetSurvey is a verb and ideally class name should be Noun..
so Survey is easier and GET/PUT can go in here...

met-api/src/met_api/models/survey.py
tiagograf-AOT: updated all models to follow same pattern

met-api/src/met_api/resources/survey.py
tiagograf-AOT: Updated all resources

met-api/migrations/versions/fc570c0faace_survey_migration.py
tiagograf-AOT: Primary key ix should be sufficient, added ix for name

djnunez-aot (approved):

Issue #293, opened 2022-07-07
Admin - Publish an Engagement


tiagograf-AOT:
![Screen Shot 2022-07-07 at 2.10.43 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/40599e59-8bf8-4ca6-bb7a-3658033a13dd)

**As an** admin

**I want** to publish an Engagement 

**so that** it can be accessed by the public

### Scenario 1: 

Admin preview the engagement/survey and click the Schedule/Publish button
Admin sees modal with the following:
Header: Schedule/Publish Engagement
Radio buttons:
<b>Publish Engagement now</b>
(The Engagement will be visible but the survey won’t be accessible until the Engagement Date.)
<b>Schedule Engagement</B>
(The Engagement will be visible on the date selected but the survey won’t be accessible
until the Engagement Date.)
Buttons: "Cancel" & "Submit"

### Scenario 2:
Admin Selects Publish Engagement now, clicks "Submit"
Admin sees a notification the Engagement was successfully published.
Modal closes.
Preview Engagement page now shows header: "Preview Engagement - Published"
"Scheduled/Published" button is replaced with "Close Preview" button (links to the Engagement listing)
Listing page now shows the engagement with the status of "Published"

Matches wireframes

https://xd.adobe.com/view/e1eae92a-d831-43c9-aaf9-4598ff2926c4-b4b9/screen/370dce46-bffc-486a-be89-b4bd4579bf97
https://xd.adobe.com/view/e1eae92a-d831-43c9-aaf9-4598ff2926c4-b4b9/screen/65d4730e-c82b-4596-835b-82938bc80cab
https://xd.adobe.com/view/e1eae92a-d831-43c9-aaf9-4598ff2926c4-b4b9/screen/f8f76aa0-3f2c-4943-afdb-6d6b702a6ee2

tiagograf-AOT:
As a User Admin I would like to publish a draft engagement so it is visible to the public

carlygreenehill:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/293?planning-poker) with ZenHub @saravanpa-aot

carlygreenehill:
[...]

PR #322, opened 2022-07-18
Publish Engagement

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/293

*Description of changes:*
Add Preview Engagement banner when viewing an engagement.
Add feature to publish engagements.
Add checks for published/draft engagements.

Screen Shot 2022-07-18 at 1 02 20 PM

### DIFF
--- a/met-api/src/met_api/models/engagement.py
+++ b/met-api/src/met_api/models/engagement.py
@@ -3,6 +3,7 @@
 Manages the engagement
 """
 from datetime import datetime
+from met_api.constants.status import Status
 from sqlalchemy.sql.schema import ForeignKey
 from sqlalchemy.dialects.postgresql import JSON
 from met_api.schemas.engagement import EngagementSchema
@@ -40,10 +41,31 @@ class Engagement(db.Model):
         return engagement_schema.dump(data)
 
     @classmethod
+    def get_published_engagement(cls, engagement_id):
+        """Get an engagement."""
+        engagement_schema = EngagementSchema()
+        data = db.session.query(Engagement)\
+            .filter(Engagement.status_id != Status.Draft.value)\
+            .filter_by(id=engagement_id)\
+            .first()
+        return engagement_schema.dump(data)
+
+    @classmethod
     def get_all_engagements(cls):
         """Get all engagements."""
         engagements_schema = EngagementSchema(many=True)
         data = db.session.query(Engagement).join(EngagementStatus).order_by(Engagement.id.asc()).all()
+        return engagements_schema.dump(data)
+
+    @classmethod
+    def get_published_engagements(cls):
+        """Get all open engagements."""
+        engagements_schema = EngagementSchema(many=True)
+        data = db.session.query(Engagement)\
+            .join(EngagementStatus)\
+            .filter(Engagement.status_id != Status.Draft.value)\
+            .order_by(Engagement.id.asc())\
+            .all()
         return engagements_schema.dump(data)
 
     @classmethod
@@ -55,7 +77,7 @@ class Engagement(db.Model):
             rich_description=engagement.get('rich_description', None),
             start_date=engagement.get('start_date', None),
             end_date=engagement.get('end_date', None),
-            status_id=1,  # Draft status
+            status_id=Status.Draft.value,
             created_by=engagement.get('created_by', None),
             created_date=datetime.utcnow(),
             updated_by=engagement.get('updated_by', None),
@@ -78,6 +100,8 @@ class Engagement(db.Model):
             rich_description=engagement.get('rich_description', None),
             start_date=engagement.get('start_date', None),
             end_date=engagement.get('end_date', None),
+            status_id=engagement.get('status_id', None),
+            published_date=engagement.get('published_date', None),
             updated_date=datetime.utcnow(),
             updated_by=engagement.get('updated_by', None),
             banner_filename=engagement.get('banner_filename', None),
--- a/met-api/src/met_api/models/engagement.py
+++ b/met-api/src/met_api/models/engagement.py
@@ -3,9 +3,9 @@
 Manages the engagement
 """
 from datetime import datetime
-from met_api.constants.status import Status
 from sqlalchemy.sql.schema import ForeignKey
 from sqlalchemy.dialects.postgresql import JSON
+from met_api.constants.status import Status
 from met_api.schemas.engagement import EngagementSchema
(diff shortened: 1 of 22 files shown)


met-web/src/components/engagement/view/ActionContext.tsx
```diff
@@ -63,6 +67,20 @@ export const ActionProvider = ({ children }: { children: JSX.Element | JSX.Eleme
     });
     const [engagementLoading, setEngagementLoading] = useState(true);
 
+    const publishEngagement = async (engagement: Engagement): Promise<Engagement> => {
+        try {
+            const result = await putEngagement(engagement);
+            //TODO engagement update success message in notification module
```
jadmsaadaot, line 73: this Todo has been done by the line after

met-web/src/components/survey/building/index.tsx
```diff
@@ -109,6 +110,17 @@ const SurveyFormBuilder = () => {
         return <FormBuilderSkeleton />;
     }
 
+    const hasEngagement = !!savedSurvey.engagement;
+    const isEngagementDraft = savedSurvey.engagement?.status_id == EngagementStatus.Draft;
+
+    if (hasEngagement && !isEngagementDraft) {
```
jadmsaadaot, line 116: Cool. I think the react pattern is to add stuff like this in a  useEffect which kinda is like the hook version of ComponentDidMount life cycle methods. You want to run that if the component mounted and depending on the savedSurvey field so I think useEffect(() => {... that block}, [savedSurvey])

jadmsaadaot:
Noticed a failing check

met-api/src/met_api/models/engagement.py
```diff
@@ -39,13 +40,34 @@ def get_engagement(cls, engagement_id):
         data = db.session.query(Engagement).filter_by(id=engagement_id).first()
         return engagement_schema.dump(data)
 
+    @classmethod
+    def get_published_engagement(cls, engagement_id):
+        """Get an engagement."""
+        engagement_schema = EngagementSchema()
+        data = db.session.query(Engagement)\
+            .filter(Engagement.status_id != Status.Draft.value)\
+            .filter_by(id=engagement_id)\
```
saravanpa-aot, line 49: you can simplify thiis with an and ;no need to of separate filter_by

met-api/src/met_api/models/engagement.py
```diff
@@ -39,13 +40,34 @@ def get_engagement(cls, engagement_id):
...
     @classmethod
     def get_all_engagements(cls):
         """Get all engagements."""
         engagements_schema = EngagementSchema(many=True)
         data = db.session.query(Engagement).join(EngagementStatus).order_by(Engagement.id.asc()).all()
         return engagements_schema.dump(data)
 
+    @classmethod
+    def get_published_engagements(cls):
+        """Get all open engagements."""
```
saravanpa-aot, line 62: I generally like get_engagement_status and pass the status 
In the model level , it will be less code for different use cases
but for readability at service , get_published_engagement will be good

saravanpa-aot (approved):

met-web/src/components/engagement/view/PreviewBanner.tsx
```diff
@@ -0,0 +1,88 @@
...
+    return (
+        <ConditionalComponent condition={isDraft && isLoggedIn}>
+            <Box
+                sx={{
+                    backgroundColor: 'secondary.light',
+                }}
+            >
+                <Grid
+                    container
+                    direction="column"
```
jadmsaadaot, line 54: This is obviously looking at the snapshot already working. I just have a feeling here that you might not need the direction='column', row perhaps would be better. The grid props like xs={12} will work only for a row direction since it is a 12 column grid. If you needed spacing between the grid items you can always add spacing={1} for the container or rowSpacing={1}

met-web/src/components/survey/building/index.tsx
tiagograf-AOT: Noted & Changed, thanks!

met-web/src/components/engagement/view/ActionContext.tsx
tiagograf-AOT: Removed

met-api/src/met_api/resources/survey.py
```diff
@@ -58,11 +58,20 @@ class Survey(Resource):
...
     # @TRACER.trace()
     @cross_origin(origins=allowedorigins())
+    @auth.requireOptional
     def get(survey_id):
         """Fetch a single survey matching the provided id."""
         try:
-            survey_record = SurveyService().get(survey_id)
-            return ActionResult.success(survey_id, survey_record)
+            user_id = TokenInfo.get_id()
+            if user_id:
```
jadmsaadaot, line 66: Trying to think of a cleaner way than having a flag in the method that decides which method is called. Maybe we can have that logic in the service at least. The resource shouldn't care about that particular business logic

jadmsaadaot: We can call one service method and it will be a small concise method def get_survey(survey_id, user_id) {

if(user_id) return get(survey_id)

return get_open(survey_id).

This way the resource doesn't have to manage this logic

met-web/src/components/survey/building/index.tsx
```diff
@@ -131,7 +143,11 @@ const SurveyFormBuilder = () => {
             </Grid>
             <Grid item xs={12}>
                 <Stack direction="row" spacing={2}>
-                    <Button variant="contained" disabled={!formData} onClick={handleSaveForm}>
+                    <Button
+                        variant="contained"
+                        disabled={!formData || (hasEngagement && !isEngagementDraft)}
```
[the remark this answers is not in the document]
tiagograf-AOT: Updated

met-api/src/met_api/resources/survey.py
tiagograf-AOT: I'll make this change but it also needs to be changed in the future, I mentioned this before but the right approach would be to have separate API endpoints like "public" and "internal" and not a common endpoint that will do both!

met-api/src/met_api/models/engagement.py
tiagograf-AOT: updated

djnunez-aot (approved):

saravanpa-aot (approved):

jadmsaadaot (approved):

PR #326, opened 2022-07-20
Add preview published engagement banner

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/293

*Description of changes:*
Adds preview published engagement banner
Fix typo


djnunez-aot (approved):

met-web/src/components/engagement/view/PreviewBanner.tsx
```diff
@@ -10,50 +10,43 @@ export const PreviewBanner = () => {
...
-            return engagement;
         }
     };
 
+    if (engagementLoading) {
+        return <Skeleton variant="rectangular" width="100%" height="10em" />;
+    }
+
     return (
-        <ConditionalComponent condition={isDraft && isLoggedIn}>
```
jadmsaadaot, line 46: if the whole component is conditional, might we consider returning early. if(isLoggedIn) return null above the long return one. This way the main logic of the component wouldn't have to be nested under this conditional component, reader would focus on its main logic.

met-web/src/components/engagement/view/PreviewBanner.tsx
```diff
@@ -10,50 +10,43 @@ export const PreviewBanner = () => {
...
-        };
-
-        return Object.values(errors).some((isError: unknown) => isError);
-    };
+    useEffect(() => {
+        setIsDraft(savedEngagement.status_id === EngagementStatus.Draft);
+    }, [savedEngagement]);
 
     const handlePublishEngagement = async () => {
-        const hasErrors = validate();
```
jadmsaadaot, line 29: can we return early instead of if else. if(!isDraft) navigate('/'); return;) then the rest of the code not under an if.  Less cognitive complexity this way

met-web/src/components/engagement/view/PreviewBanner.tsx
```diff
@@ -64,15 +57,15 @@ export const PreviewBanner = () => {
...
                         >
                             <Button variant="outlined" onClick={() => navigate(`/engagement/form/${engagementId}`)}>
-                                Edit Enagagement
+                                Edit Engagement
                             </Button>
                         </Box>
                         <Button
                             variant="contained"
                             sx={{ marginLeft: '1em' }}
                             onClick={() => handlePublishEngagement()}
```
jadmsaadaot, line 66: we can maybe consider changing the name here. Since this button either publishes or closes preview. So handlePublishEngagement to do both of those things could be confusing

met-web/src/components/engagement/view/PreviewBanner.tsx
```diff
@@ -10,50 +10,43 @@ export const PreviewBanner = () => {
...
 
-    const validate = () => {
-        const errors = {
-            status: savedEngagement.status_id !== EngagementStatus.Draft,
-        };
-
-        return Object.values(errors).some((isError: unknown) => isError);
-    };
+    useEffect(() => {
+        setIsDraft(savedEngagement.status_id === EngagementStatus.Draft);
```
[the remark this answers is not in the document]
tiagograf-AOT: I had to use useEffect for it to work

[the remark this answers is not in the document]
jadmsaadaot: I see, weird thought because we are already using this pattern somewhere else and it's working. Like const isNewEngagement = engagementId === 'create';  in EngagementForm. The setIsDraft needs to be in a useEffect for sure. But a simple const would still work no? Would you mind trying again just using const isDraft = savedEngagement.status_id === EngagementStatus.Draft?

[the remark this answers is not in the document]
tiagograf-AOT: It worked!

met-web/src/components/engagement/view/PreviewBanner.tsx
tiagograf-AOT: Updated

met-web/src/components/engagement/view/PreviewBanner.tsx
tiagograf-AOT: Updated

met-web/src/components/engagement/view/PreviewBanner.tsx
tiagograf-AOT: Updated

Issue #234, opened 2022-06-17
As an Admin, I want to select an existing survey to add to an Engagement


marieanik:
**As an** Admin

**I want** to select an existing survey 

**so that** add it to my engagement

### Scenario 1: 
Admin clicks the "Create Survey" button on the engagement page.

Admin sees the "Create Survey" screen. 

Section Header: "Create Survey" (+ edit icon = widget name is editable)

Radio buttons for the following options:
Create a new Survey
Clone an existing Survey
Add an existing Survey to my Engagement

"Save & Continue" and "Cancel" buttons

### Scenario 2:
Admin selects "Add an existing Survey to my Engagement"
Selects Survey from drop-down, (only showing surveys not already attached to an engagement)
"Save & Continue" and "Cancel" buttons
User clicks "Save & Continue" , goes back to Engagement page with survey added in the survey section

### Scenario 3:
Matches wireframes
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/bf0d9afb-1c4b-4ab3-a08f-8ed5a7b35163/
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/99c014f2-c818-4552-932a-6d4a912c0092
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/f2b3759e-bacb-41ae-ae84-e74c304877d8/

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

Issue #47, opened 2022-04-27
An Admin User can create a survey directly on the engagement page to add to my engagement 


sidtobias:
**As an** Admin

**I want** to create a survey 

**so that** I can to add it to my engagement

### Scenario 1: 

Admin is on the Create/Edit engagement page and Clicks the button "Add Survey" in the Surveys section.

### Scenario 2:

Admin sees the "Create Survey" screen.

Section Header: "Create Survey" (+ edit icon = Survey name is editable)
Radio buttons for the following options:
Create a new Survey
Clone an existing Survey
Add an existing Survey to my Engagement

"Save & Continue" and "Cancel" buttons 

Admin selects "Create a new Survey", sees the input field "Enter Survey Name" enters Survey Name and clicks on "Save & Continue" button. Survey page opens. Survey is created in database and show on Engagement page & Survey listing page.

### Scenario 3:
Confirm user interface to the visual design
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/d5ff35b0-90ea-49eb-a8f2-fe307a156b92
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/bf0d9afb-1c4b-4ab3-a08f-8ed5a7b35163/
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/c180802e-cc2d-4bde-9ae6-231dcce70e65
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/f2b3759e-bacb-41ae-ae84-e74c304877d8/
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/2a77ed82-38d7-4a24-be8a-f6d329a807ed/

### Validation Rules? 

If admin clicks the button to add a survey (or widget), Engagement must contain at least the engagement name and be saved. 

Validation message "Please add the Engagement Name and save the Engagement before adding a Survey."

### Screenshot 

(Link above will always be the most up-to-date version.)

sidtobias:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/47?planning-poker) with ZenHub @djnunez-aot @jadmsaadaot @nikhila-aot @saravanpa-aot @tiagograf-AOT

[...]

PR #329, opened 2022-07-20
Add existing survey to an engagement

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/234
https://github.com/bcgov/met-public/issues/47

*Description of changes:*

- Add new options in survey page to add existing survey
- Add new endpoint to link survey to engagement
- Add API calling method in survey service to call link engagement
- Add replaceAllInUrl helper method to replace placeholder path variables with values


djnunez-aot (approved):

met-api/src/met_api/models/survey.py
```diff
@@ -56,6 +56,13 @@ def get_all_surveys(cls) -> List[SurveySchema]:
         data = db.session.query(Survey).join(Engagement, isouter=True).join(EngagementStatus, isouter=True).all()
         return survey_schema.dump(data)
 
+    @classmethod
+    def get_all_unlinked_surveys(cls) -> List[SurveySchema]:
+        """Get all surveys that are unlinked to engagement."""
+        survey_schema = SurveySchema(many=True)
+        data = db.session.query(Survey).filter_by(engagement_id=None).all()
```
saravanpa-aot, line 63: just a suggestion , call it surveys instead of data ...easily comprehensible that way

met-api/src/met_api/models/survey.py
```diff
@@ -89,3 +96,17 @@ def update_survey(cls, survey: SurveySchema) -> DefaultMethodResult:
...
+        """Update engagement."""
+        update_fields = dict(
+            engagement_id=engagement_id,
+        )
+        query = Survey.query.filter_by(id=survey_id)
+        record = query.first()
+        if not record:
+            return DefaultMethodResult(False, 'Survey Not Found', survey_id)
+        query.update(update_fields)
+        db.session.commit()
```
saravanpa-aot, line 111: I think if you just do survey.engagement_id = eng_id and session.commit ;it will all work...might increase the readability ..

met-api/src/met_api/resources/survey.py
```diff
@@ -132,3 +133,28 @@ def put():
             return ActionResult.error(str(err))
         except ValueError as err:
             return ActionResult.error(str(err))
+
+
+@cors_preflight('PUT,OPTIONS')
+@API.route('/<survey_id>/link/engagement/<engagement_id>')
+class SurveyLink(Resource):
```
saravanpa-aot, line 140: hm... generally its just a put on survey/:id ; we dont need the link action in the rest call ... 
but we can treat this as an exception since linking is a major thing...
normally we dont need to write multiple rest calls for each action in the UI..
call put on the suveys/:id and see if the request json has engId ;if so update it...

met-api/src/met_api/models/survey.py
```diff
@@ -89,3 +96,17 @@ def update_survey(cls, survey: SurveySchema) -> DefaultMethodResult:
         query.update(update_fields)
         db.session.commit()
         return DefaultMethodResult(True, 'Survey Updated', survey_id)
+
+    @classmethod
+    def link_survey(cls, survey_id, engagement_id) -> DefaultMethodResult:
+        """Update engagement."""
```
tiagograf-AOT, line 102: wrong docstring

met-web/src/helper/index.ts
```diff
@@ -2,4 +2,15 @@ const replaceUrl = (URL: string, key: string, value: string) => {
...
+    [param: string]: string;
+}
+const replaceAllInURL = ({ URL, params }: { URL: string; params: Params }) => {
+    const regex = new RegExp(Object.keys(params).join('|'), 'gi');
+
+    return URL.replace(regex, function (matched) {
+        return params[matched];
+    });
+};
+
```
saravanpa-aot, line 15: it's just my cynical mind. does it save a lot of code ? can we achieve this by normal variable replacement or something like that?
asking bcoz we dont have any test to cover this method and if some point , some bug happens, will be hard to debug this..
if we have a strong use case ;like replacing lot of variables in url always , i am all in for it..
but just worried about maintainability...if its just replacing our params in url , may be simple variable substitions..
lol..definitly point of contention..

met-web/src/components/survey/create/OptionsFormSkeleton.tsx
```diff
@@ -0,0 +1,39 @@
...
+                xs={12}
+                container
+                direction="row"
+                justifyContent="flex-start"
+                alignItems="center"
+                spacing={1}
+            >
+                <Grid item xs={12}>
+                    <Typography variant="h6">
+                        <Skeleton />
```
saravanpa-aot, line 25: curious , whats this skeleton?

met-api/src/met_api/services/survey_service.py
```diff
@@ -33,12 +33,25 @@ def supply_banner_url(engagement):
         engagement['banner_url'] = ObjectStorageService.get_url(engagement.get('banner_filename', None))
         return engagement
 
+    @classmethod
+    def get_surveys(cls, unlinked=False):
```
tiagograf-AOT, line 37: I don't get the optional parameter here, isn't it better to just call the method?

met-web/src/components/survey/create/CreateSurveyContext.tsx
```diff
@@ -21,7 +37,35 @@ interface SurveyForm {
...
+    const location = useLocation();
+    const searchParams = new URLSearchParams(location.search);
+    const engagementId = searchParams.get('engagementId');
+
+    useEffect(() => {
+        fetchEngagementToLink();
+    }, [engagementId]);
+
+    const fetchEngagementToLink = async () => {
+        if (!engagementId || isNaN(Number(engagementId))) {
```
saravanpa-aot, line 56: !engagementId || isNaN(Number(engagementId))
looks like a complicated check... do we need the isNan and all?

saravanpa-aot (approved):
Large PR sir ..if possible please split into backend and frontend.. took a lot of energy  to review the whole thing

met-api/src/met_api/resources/survey.py
jadmsaadaot: That was my initial way. However, when I implemented that and allowed for engagement_id to be updated in the model class for the update method. I had to put checks for when updating the survey (say adding components) to not update engagement_id with None and lose the link. When I saw that, I felt uncomfortable using the update method to link and to update survey components. I felt link should be on its own.

met-web/src/helper/index.ts
jadmsaadaot: Some URLs may have multiple path variables, I added the above to handle survey/survey_id/engagement/engagement_id. I I thought we might need a method to handle that in one go. Researched and took it from here https://stackoverflow.com/questions/15604140/replace-multiple-strings-with-multiple-other-strings. 
What do you think? Not necessary?
I thought if the helper method itself is ugly, it would at least be hidden and it's calling is clear and easy.

met-web/src/components/survey/create/OptionsFormSkeleton.tsx
jadmsaadaot: It's a cool MUI component to have a loading animation. Instead of having a circular loader. Check it out here https://mui.com/material-ui/react-skeleton/

met-web/src/components/survey/create/CreateSurveyContext.tsx
jadmsaadaot: Just in case someone played with the URL and added '$%^&*' for engagementId

met-api/src/met_api/resources/survey.py
saravanpa-aot: hm..since linking is major thing for us ,its okay to have like this..but if its a smaller operation , we should try to accomdate in the Put itself..
yeah..points agreed..we have to validate the None and all..More code for sure

met-web/src/apiManager/endpoints/index.ts
```diff
@@ -17,6 +17,7 @@ const Endpoints = {
         GET_ALL: `${AppConfig.apiUrl}/survey/`,
         CREATE: `${AppConfig.apiUrl}/survey/`,
         UPDATE: `${AppConfig.apiUrl}/survey/`,
+        LINK_TO_ENGAGEMENT: `${AppConfig.apiUrl}/survey/survey_id/link/engagement/engagement_id`,
```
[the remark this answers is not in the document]
jadmsaadaot: Agreed, I didn't know what's best to do here. The replaceAllInUrl method kinda works by having the keys as the placeholders like the above (survey_id, engagement_id) and characters like  < and > aren't allowed in keys.  What do you think? Should we stick to that standard?

met-api/src/met_api/services/survey_service.py
jadmsaadaot: How do you mean just call the method?

I need to either get all surveys or just the unlinked ones.

met-api/src/met_api/models/survey.py
jadmsaadaot: noted, will follow this up with change in a new PR

met-api/src/met_api/models/survey.py
jadmsaadaot: noted, will do

met-api/src/met_api/models/survey.py
jadmsaadaot: noted will do

PR #353, opened 2022-07-27
Unit Testing

djnunez-aot:
*Issue #:*

*Description of changes:*


met-web/src/config.ts
```diff
@@ -20,24 +20,22 @@ declare global {
     }
 }
 
-const API_URL = window._env_.REACT_APP_API_URL || process.env.REACT_APP_API_URL;
```
tiagograf-AOT, line 23: I'm worried that this would break our existing way of loading environment variables in our DEV/TEST environments

met-web/tests/unit/components/setEnvVars.tsx
```diff
@@ -0,0 +1,6 @@
+export function setupEnv() {
+    process.env.REACT_APP_KEYCLOAK_CLIENT = 'met-web';
```
tiagograf-AOT, line 2: I think we may need to tweak this in the future

tiagograf-AOT (dismissed):
I have a few concerns with the environment variable stuff, but other than that good job!

met-web/src/config.ts
djnunez-aot: you might be right, might need a little help setting something up for the testing then

tiagograf-AOT: I did check and thats how we load the environment variables, can you somehow pass an empty object to the window._env_ when loading the tests?
e.g.
window['_env_'] = {};

met-web/tests/unit/components/setEnvVars.tsx
djnunez-aot: any suggestions?

met-web/src/config.ts
tiagograf-AOT: we do this in the public/config.js file

djnunez-aot: i added the config.js to the jest setup env so and added the window.env back into the file so everything should be good now :)

tiagograf-AOT (approved):

jadmsaadaot (approved):

Issue #379, opened 2022-08-03
Submission fails if comments field is empty


tiagograf-AOT:
Actual
When submitting a survey, if an optional text field is left empty, the submission fails
Expected
When submitting a survey, if an optional text field is left empty, the submission should work

PR #388, opened 2022-08-05
Submission bug

djnunez-aot:
*Issue #379:*
https://github.com/bcgov/met-public/issues/#379


tiagograf-AOT:
It doesn't look like this would solve the problem, were you able to reproduce the issue?

met-api/src/met_api/services/comment_service.py
```diff
@@ -64,20 +64,19 @@ def extract_comments(cls, survey_submission: SubmissionSchema, survey: SurveySch
...
         if len(components) == 0:
             return []
-
         # get the 'key' for each component that has 'inputType' text and filter out the rest.
         comments_keys = [
             component.get('key', None) for component in components
             if component.get('inputType', None) == 'text']
-
         submission = survey_submission.get('submission_json', {})
         comments_texts = [submission.get(key, None) for key in comments_keys]
```
tiagograf-AOT, line 72: I'm gonna throw a different approach and leave it to Jad to decide, but when assigning comments_texts you can filter out the blank ones something like:  
        comments_texts = [submission.get(key, None) for key in comments_keys
        if submission.get(key, None) != '']

tiagograf-AOT (approved):

met-api/src/met_api/services/comment_service.py
jadmsaadaot: yup this is definitely cleaner and more performant

jadmsaadaot (changes requested):
Please filter empty texts Tiago's way

met-api/src/met_api/services/comment_service.py
djnunez-aot: good idea

jadmsaadaot (approved):

Issue #354, opened 2022-07-27
Tweak to buttons to match B.C. Gov Style


marieanik:
Change font-size to 1.125 rem (My bad Jad, told you 1.2 but it is 1.125)

Primary (blue) button:
Hover is same blue #003366 but with opacity 0.8 and text underlined

.BC-Gov-PrimaryButton:hover {
  text-decoration: underline;
  opacity: 0.80;
}

https://developer.gov.bc.ca/Design-System/Primary-Button

Secondary (white) button:

Border is 2px

Hover is blue #003366, opacity 0.8, white text, underlined

.BC-Gov-SecondaryButton:hover {
    opacity: 0.80;
    text-decoration: underline;
    background-color: #003366;
    color: #FFFFFF;
    }

https://developer.gov.bc.ca/Design-System/Secondary-Button

PR #431, opened 2022-08-11
Make primary and secondary buttons BV Gov standard

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/354

*Description of changes:*

- Update hover effect and text transform for primary and secondary buttons


djnunez-aot (approved):

tiagograf-AOT (approved):
Looks good! Just a nit but I don’t think we need to pass the variant to the primary /secondary buttons anymore, can you move that to in the custom component?!

Issue #412, opened 2022-08-09
Update Survey table "Comments" header to "Responses" to match the Engagement table


marieanik:
It was decided that "Responses" was preferable and this made it to the Engagement table. Let's change it on the survey page too.

Issue #410, opened 2022-08-09
Tweak Engagement listing table


marieanik:
Remove N/A and leave filed blank when there is nothing to show.

Remove the View Report link until there is at least one response when the Engagement is open (until there is something to see in the report).

![Screen Shot 2022-08-09 at 2.26.40 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/1e98393c-3c6b-426a-9be3-a35ccc3a8c2d)

PR #440, opened 2022-08-15
Table Fixes 

djnunez-aot:
*Issue #410 ,#412*

Small Table fixes

### DIFF
--- a/met-web/src/components/LandingPage/LandingPage.tsx
+++ b/met-web/src/components/LandingPage/LandingPage.tsx
@@ -80,7 +80,7 @@ const LandingPage = () => {
             allowSort: false,
             getValue: (row: Engagement) => {
                 if (row.surveys.length === 0) {
-                    return 'N/A';
+                    return '';
                 }
 
                 return (
@@ -98,7 +98,7 @@ const LandingPage = () => {
             allowSort: false,
             getValue: (row: Engagement) => {
                 if (row.surveys.length === 0) {
-                    return 'N/A';
+                    return '';
                 }
 
                 return row.surveys[0].responseCount;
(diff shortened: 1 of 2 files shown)


met-web/src/components/LandingPage/LandingPage.tsx
```diff
@@ -112,13 +112,19 @@ const LandingPage = () => {
...
                 if (row.surveys.length === 0) {
-                    return 'N/A';
+                    return '';
                 }
 
                 return (
-                    <MuiLink component={Link} to={`/survey/${Number(row.surveys[0].id)}/results`}>
-                        View Report
-                    </MuiLink>
+                    <>
```
jadmsaadaot, line 119: The above early retun is there in case we need not to show View Report link. Early return was used because it's cleaner and clearer. Why not use it for when row.surveys[0].responseCount === 0 too

jadmsaadaot (changes requested):
Use the already existing early return to hide a cell

met-web/src/components/LandingPage/LandingPage.tsx
```diff
@@ -111,8 +111,8 @@ const LandingPage = () => {
             label: 'Reporting',
             allowSort: false,
             getValue: (row: Engagement) => {
-                if (row.surveys.length === 0) {
-                    return 'N/A';
+                if (row.surveys.length === 0 || row.surveys[0].responseCount === 0) {
```
jadmsaadaot, line 114: is survey.responseCount ever populated with data from the API?

jadmsaadaot (changes requested):
Is Survey.ResponseCount ever populated from the API?

PR #442, opened 2022-08-15
Etl refactor

saravanpa-aot:
*Issue #:*

*Description of changes:*


tiagograf-AOT (approved):

VineetBala-AOT (approved):

met-cron/src/met_cron/utils/__init__.py
```diff
@@ -19,6 +19,17 @@ class FormIoComponentType(Enum):
...
     TEXT = 'text'
-    SIMPLE_TEXT = 'simpletextfield'
+    SIMPLE_TEXTFIELD = 'simpletextfield'
+    TEXTFIELD = 'textfield'
+    SIMPLE_TEXT_AREA = 'simpletextarea'
+    TEXT_AREA = 'textarea'
+
+
+RADIO_TYPES = (FormIoComponentType.SIMPLE_CHECKBOXES.value,)
+SELECTBOXES_TYPES = (FormIoComponentType.SIMPLE_CHECKBOXES.value, FormIoComponentType.CHECKBOX.value, FormIoComponentType.SELECTBOXES.value)
```
jadmsaadaot, line 33: I don't think we need this anymore. In formio there is inputType as part of the schema, we can specify there what we want. Select boxes and checkboxes already have input type 'checkbox' and if we need to specify like here https://github.com/bcgov/met-formio/blob/main/src/components/SimpleTextField/Component.ts we can.

So  grouping like RADIO_TYPES  is not necessary anymore as we can have that grouping by specifying several components with same inputType if you want that.

jadmsaadaot: Postal code component, this is a custom inputType I specified

https://github.com/bcgov/met-formio/blob/main/src/components/SimplePostalCode/Component.ts

saravanpa-aot: Good Point...I thought thats the case..
But just wanted ETL logic to work with some flexibility..
once we think through all the custom components , we can come back and change this I think..
Is it okay to merge as is? And good side is this pattern is there and its for next developer to understand that we can target multiple items in a line
Also DEV and TEST has lot of test data with all these values (probably old data) ;so helps me with testing

saravanpa-aot: I havent tested postal code yet..wildo

jadmsaadaot (approved):

PR #444, opened 2022-08-15
Remove dashboard page reliance on engagement view context

jadmsaadaot:
*Issue #:*

*Description of changes:*

- Make dashboard component independent of Engagement View component

### DIFF
--- a/met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
+++ b/met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
@@ -28,7 +28,7 @@ export const EngagementDashboard = () => {
         const neverOpened = engagementToValidate?.submission_status < SubmissionStatus.Open;
 
         if (neverOpened) {
-            throw new Error('Engagement has not been yet opened');
+            throw new Error('Engagement has not yet been opened');
         }
     };
     useEffect(() => {
--- a/met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
+++ b/met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
@@ -25,10 +25,10 @@ export const EngagementDashboard = () => {
     const [isEngagementLoading, setEngagementLoading] = useState(true);
 
     const validateEngagement = (engagementToValidate: Engagement) => {
-        const isClosed = engagementToValidate?.submission_status === SubmissionStatus.Closed;
+        const isOpen = engagementToValidate?.submission_status === SubmissionStatus.Open;
 
-        if (!isClosed) {
-            throw new Error('Engagement is not yet closed');
+        if (!isOpen) {
+            throw new Error('Engagement is not yet open');
         }
     };
     useEffect(() => {
--- a/met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
+++ b/met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
@@ -1,92 +1,124 @@
-import React, { useContext } from 'react';
-import { Grid } from '@mui/material';
-import { useAppSelector } from 'hooks';
-import { Link } from 'react-router-dom';
-import { useNavigate } from 'react-router';
-import { ActionContext } from '../../view/ActionContext';
-import { EngagementBanner } from '../../view/EngagementBanner';
-import { PreviewBanner } from '../../view/PreviewBanner';
-import { SubmissionStatus } from 'constants/engagementStatus';
+import React, { useEffect, useState } from 'react';
+import { Grid, Link as MuiLink, Skeleton, Typography } from '@mui/material';
+import { Link, useNavigate, useParams } from 'react-router-dom';
 import { AppConfig } from 'config';
-import { PrimaryButton } from 'components/common';
+import { MetPaper, PrimaryButton } from 'components/common';
+import { ReportBanner } from './ReportBanner';
+import { createDefaultEngagement, Engagement } from 'models/engagement';
+import { SubmissionStatus } from 'constants/engagementStatus';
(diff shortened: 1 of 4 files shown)


VineetBala-AOT (approved):

jadmsaadaot:
I can't see a way around the duplication for now, unless we refactor how we create the engagement view pages

tiagograf-AOT (approved):

met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
```diff
@@ -1,92 +1,124 @@
...
-    const isOpen = savedEngagement.submission_status === SubmissionStatus.Open;
 
-    const handleStartSurvey = () => {
-        if (!isPreview) {
-            return;
+    const [engagement, setEngagement] = useState<Engagement>(createDefaultEngagement());
+    const [isEngagementLoading, setEngagementLoading] = useState(true);
+
+    const validateEngagement = (engagementToValidate: Engagement) => {
+        const neverOpened = engagementToValidate?.submission_status < SubmissionStatus.Open;
```
saravanpa-aot, line 28: whats the logic here?
the less than comparison might deserve a comment to explain it..

met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
```diff
@@ -1,92 +1,124 @@
...
+
+        if (neverOpened) {
+            throw new Error('Engagement has not yet been opened');
```
saravanpa-aot, line 31: what happens in the UII when we throw an error?

saravanpa-aot (approved):
left some comments ;plz check
good work

met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
jadmsaadaot: This method is called under a try catch block. The error should be caught and this should run:

catch (error) {
                console.log(error);
                dispatch(
                    openNotification({
                        severity: 'error',
                        text: getErrorMessage(error) || 'Error occurred while fetching Engagement information',
                    }),
                );
                navigate('/');
            }

met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
jadmsaadaot: I think @tiagograf-AOT made a point somewhere that there might be statuses other than pending open close (like draft) and these pages shouldn't be accessible if status was pending or draft, so open should have an id greater than id of pending and draft. so if engagementToValidate?.submission_status less than Open (i.e. pending or draft) it shouldn't be displayed

jadmsaadaot: As for clarifying the logic, the neverOpened variable name clears it out no?

saravanpa-aot: an explicit status check is much better..
very risky if id has to change in backend for some reason..I wouldn't expect the id has a meaning in terms of the values..
say we have to introduce a new status which belongs to open case, the new id will be more than that..so this logic will fail ..

met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
saravanpa-aot: hm..got it now.. 
I Would try to rewrite it a little bit ;but its alrite as it is..
Since the method is named as validateEngagement , exception raising is not  necessarily needed.. We are here using an exception to control the flow which is not ideal.. it can be a isValidEng or isOpenEngagement which returns a boolean and then u can do ur logic..isnt it more readable that way?
But definitely this is a well written code..so u ignore me for being picky

jadmsaadaot: It seems you prefer having an if statement to check an isValid boolean over handling validation via exception. This way is more straightforward sure, but I don't think it adheres better to the Open for extension Closed for change principle.

The try catch you can just extend the code if you needed another validation or check step.  You only have to write that method and call it inside the try.

But the boolean way, if you have to add another validation you will have to change your code to handle that boolean. Which violates the open close. I prefer the try catch and I actually think it's closer to the ideal

met-web/src/components/engagement/dashboard/report/EngagementDashboard.tsx
jadmsaadaot: We have removed the incrementation of the id for the status models. Every new status we add we have to manually give the id number. I ask to consider which code you prefer:

engagementToValidate?.submission_status === SubmissionStatus.Pending || engagementToValidate?.submission_status === SubmissionStatus.Draft || engagementToValidate?.submission_status === SubmissionStatus.Scheduled

or just simply engagementToValidate?.submission_status < SubmissionStatus.Open? After having sorted the ids properly on create

saravanpa-aot: engagementToValidate?.submission_status  in ( SubmissionStatus.PendingSubmissionStatus.Draft,SubmissionStatus.Scheduled)

Or I would create a constant list with a sensisble name IS_NOT_OPEN_STATUSES =  [SubmissionStatus.PendingSubmissionStatus.Draft,SubmissionStatus.Scheduled]  and check engagementToValidate?.submission_status in IS_NOT_OPEN_STATUSES

jadmsaadaot: Understood.

handled in https://github.com/bcgov/met-public/pull/448

Issue #383, opened 2022-08-04
What we heard report


tiagograf-AOT:
As a public user I want to be notified when the engagement period is over (closed) with the results of the survey so that I can learn about what were the outcomes.

Email content:

Title: What We Heard Report - <project name>.

Content: Thank you for taking the time to voice your opinion about <project name>. The Engagement period is now closed. You can <read all the comments> and <view the survey results> online.

Your comments will be passed along to the proponent, who will read and address each one over the next few weeks.

Thank you,

The EAO Team

tiagograf-AOT:
Update cron job to look for published engagements with end dates in the past, Update engagement status to close and send email to users that have answered the survey with a link to the results.

tiagograf-AOT:
@marieanik we need a template for the "What we heard" email

tiagograf-AOT:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/383?planning-poker) with ZenHub @djnunez-aot @jadmsaadaot @saravanpa-aot @VineetBala-AOT

PR #449, opened 2022-08-16
What we heard report Fixes

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/383

*Description of changes:*
Fixes notify api and what we heard report

### DIFF
--- a/met-api/src/met_api/utils/template.py
+++ b/met-api/src/met_api/utils/template.py
@@ -1,8 +1,10 @@
 """Template Services."""
 
+import os
 from jinja2 import Environment, FileSystemLoader
 
-ENV = Environment(loader=FileSystemLoader('./templates'), autoescape=True)
+templates_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../..', 'templates'))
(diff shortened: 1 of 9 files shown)


met-api/src/met_api/utils/template.py
```diff
@@ -1,8 +1,10 @@
 """Template Services."""
 
+import os
 from jinja2 import Environment, FileSystemLoader
 
-ENV = Environment(loader=FileSystemLoader('./templates'), autoescape=True)
+templates_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../..', 'templates'))
```
tiagograf-AOT, line 6: @saravanpa-aot this is back, when using the package (like cron does) it won't find the template again, this is an attempt fix, i'm not sure if it solves the problem yet

saravanpa-aot: hm..do we need this if we are doing gc notify?

saravanpa-aot (approved):

VineetBala-AOT (approved):

PR #501, opened 2022-08-24
Fix Survey SearchBar

djnunez-aot:
*Issue #500 :*
https://github.com/bcgov/met-public/issues/500

-Added search bar functionality to survey page

-Improved search bar so the table updates as the user is typing 

### DIFF
--- a/met-web/src/components/survey/listing/index.tsx
+++ b/met-web/src/components/survey/listing/index.tsx
@@ -43,11 +43,6 @@ const SurveyListing = () => {
         });
     };
 
-    const handleSearchBarInput = (engagementNameFilter: string) => {
-        setSearchText(engagementNameFilter);
-        handleSearchBarClick(engagementNameFilter);
-    };
-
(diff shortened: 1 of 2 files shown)


met-web/src/components/survey/listing/index.tsx
```diff
@@ -44,6 +43,11 @@ const SurveyListing = () => {
         });
     };
 
+    const handleSearchBarInput = (engagementNameFilter: string) => {
```
jadmsaadaot, line 46: This defeats the purpose of having a click button for the search. Since a click button is part of the design, we will assume that that is the UX needed.  So we won't trigger handleSearchBarClick on text change. We will trigger it on button click

djnunez-aot: removed

jadmsaadaot (approved):

tiagograf-AOT (approved):

Issue #404, opened 2022-08-09
Change "Survey" Header to be same size as Survey name. Make survey name editable.


marieanik:
![Screen Shot 2022-08-09 at 11.51.09 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/225003f9-72ef-4bd1-91d7-ea04de6fc126)
![Screen Shot 2022-08-09 at 11.49.11 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/0bce4b53-6eea-4b2a-a03d-5527e5050c05)
![Screen Shot 2022-08-09 at 11.47.37 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/db8cf5e5-206e-4af9-861e-4300d28d90a6)

PR #521, opened 2022-08-25
Add survey name update and change header size

djnunez-aot:
*Issue #404 :*
https://github.com/bcgov/met-public/issues/404

-Add survey name changing

-Change survey header size

### DIFF
--- a/met-api/src/met_api/models/survey.py
+++ b/met-api/src/met_api/models/survey.py
@@ -92,6 +92,7 @@ class Survey(db.Model):  # pylint: disable=too-few-public-methods
             form_json=survey.get('form_json', None),
             updated_date=datetime.now(),
             updated_by=survey.get('updated_by', None),
+            name=survey.get('name', ''),
         )
         survey_id = survey.get('id', None)
         query = Survey.query.filter_by(id=survey_id)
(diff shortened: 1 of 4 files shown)


jadmsaadaot (approved):

met-api/src/met_api/models/survey.py
```diff
@@ -92,6 +92,7 @@ def update_survey(cls, survey: SurveySchema) -> DefaultMethodResult:
             form_json=survey.get('form_json', None),
             updated_date=datetime.now(),
             updated_by=survey.get('updated_by', None),
+            name=survey.get('name', ''),
```
saravanpa-aot, line 95: name=survey.get('name', ''),

incase if UI fails to send , it might become nulll..
to avoid it , https://github.com/bcgov/met-public/pull/511 see this

saravanpa-aot (commented):
Put a comment to address the survey name being null

met-api/src/met_api/models/survey.py
djnunez-aot: added the record check to prevent the null case

saravanpa-aot (approved):

Issue #464, opened 2022-08-18
Make modal reusuable by creating a reusable notification modal


jadmsaadaot:
The first modal is for notification. The second one is to confirm an action.
![Screen Shot 2022-09-02 at 10.16.41 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/ea7421bc-d82c-45f0-ad16-2e76ee5c1770)
![Screen Shot 2022-09-02 at 10.16.54 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/342f9a03-e160-49f4-8b65-bdee8f5dd695)

jadmsaadaot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/464?planning-poker) with Zenhub @saravanpa-aot @tiagograf-AOT @VineetBala-AOT

PR #575, opened 2022-09-08
Modal Store

djnunez-aot:
*Issue #464 :*
https://github.com/bcgov/met-public/issues/464

-Created Modal Slice

-Added Modal dispatching to application

### DIFF
--- a/met-web/src/components/common/Modals/ConfirmModal.tsx
+++ b/met-web/src/components/common/Modals/ConfirmModal.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { Grid, Stack, useMediaQuery, Theme } from '@mui/material';
+import { modalStyle, PrimaryButton, SecondaryButton, MetHeader1, MetBody } from 'components/common';
+import { ModalProps } from './types';
+
+const ConfirmModal = ({ header, subTextArray, handleClose }: ModalProps) => {
+    const isSmallScreen = useMediaQuery((theme: Theme) => theme.breakpoints.down('sm'));
+    return (
+        <Grid
+            container
+            direction="row"
+            justifyContent="flex-start"
+            alignItems="space-between"
+            sx={{ ...modalStyle }}
+            rowSpacing={2}
+        >
+            <Grid container item xs={12}>
+                <Grid item xs={12}>
+                    <MetHeader1 sx={{ fontWeight: 'bold', mb: 2 }}>{header}</MetHeader1>
+                </Grid>
+            </Grid>
+            <Grid container direction="row" item xs={12}>
+                {subTextArray.map((subText: string) => (
+                    <Grid item xs={12}>
+                        <MetBody sx={{ mb: 1 }}> {subText} </MetBody>
+                    </Grid>
+                ))}
+                <Grid
+                    item
+                    container
+                    direction={{ xs: 'column', sm: 'row' }}
+                    xs={12}
+                    justifyContent="flex-end"
+                    spacing={1}
+                    sx={{ mt: '1em' }}
+                >
+                    <Stack direction={{ xs: 'column', sm: 'row' }} spacing={1} width="100%" justifyContent="flex-end">
+                        {isSmallScreen ? (
+                            <>
+                                <PrimaryButton type="submit" variant={'contained'}>
+                                    Confirm
+                                </PrimaryButton>
+                                <SecondaryButton onClick={handleClose}>Cancel</SecondaryButton>
+                            </>
+                        ) : (
+                            <>
+                                <SecondaryButton onClick={handleClose}>Cancel</SecondaryButton>
+                                <PrimaryButton type="submit" variant={'contained'}>
+                                    Confirm
+                                </PrimaryButton>
+                            </>
+                        )}
+                    </Stack>
+                </Grid>
+            </Grid>
+        </Grid>
+    );
+};
+
+export default ConfirmModal;
--- a/met-web/src/components/common/modal.tsx
+++ b/met-web/src/components/common/modal.tsx
@@ -1,0 +1,27 @@
+import * as React from 'react';
+import { useAppSelector, useAppDispatch } from '../../hooks';
+import Modal from '@mui/material/Modal';
+import { closeModal } from 'services/modalService/modalSlice';
+import CloseModal from './Modals/UpdateModal';
+import ConfirmModal from './Modals/ConfirmModal';
+
+export const ModalProvider = () => {
+    const dispatch = useAppDispatch();
+    const open = useAppSelector((state) => state.modal.open);
+    const modalData = useAppSelector((state) => state.modal.data);
+    const type = useAppSelector((state) => state.modal.type);
+
+    function handleClose() {
+        dispatch(closeModal());
+    }
+
+    return (
+        <Modal open={open} onClose={handleClose}>
+            {type === 'closeModal' ? (
+                <CloseModal header={modalData.header} subTextArray={modalData.subText} handleClose={handleClose} />
+            ) : (
+                <ConfirmModal header={modalData.header} subTextArray={modalData.subText} handleClose={handleClose} />
+            )}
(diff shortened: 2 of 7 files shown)


met-web/src/components/engagement/view/EngagementView.tsx
```diff
@@ -66,6 +67,27 @@ export const EngagementView = () => {
                     </Grid>
                     <Grid item xs={12} lg={4}>
                         <WhoIsListeningWidget />
+                        <Button
+                            onClick={() =>
```
djnunez-aot, line 71: For Testing Purposes Only

met-web/src/components/common/Modals/ConfirmModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+                    xs={12}
+                    justifyContent="flex-end"
+                    spacing={1}
+                    sx={{ mt: '1em' }}
+                >
+                    <Stack direction={{ xs: 'column', sm: 'row' }} spacing={1} width="100%" justifyContent="flex-end">
+                        {isSmallScreen ? (
+                            <>
+                                <PrimaryButton
+                                    onClick={() => buttons[0].buttonFunction}
```
jadmsaadaot, line 41: I think instead of sending an array and indexing it like this, which is a bit ambiguous. Let's do something similar to MetWidget in the common components. Notice there the handle functions are props, if they are passed we render the corresponding button and if not we hide. Let button text be standard and static. Let that not be a variable since it would make things more difficult.

met-web/src/components/common/modal.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+    return (
+        <Modal
+            sx={{ border: '2px solid red' }}
+            open={open}
+            onClose={handleClose}
+            aria-labelledby="modal-modal-title"
+            aria-describedby="modal-modal-description"
+        >
+            {type === 'closeModal' ? (
```
jadmsaadaot, line 26: Not cool, if we had many modals this would be very difficult. Maybe use <ConditionalComponent condition={type==='bla'> on top of each other

jadmsaadaot, line 26: not a cool name closeModal, the suffix Modal is unnecessary. and the verb close is not exclusive to either. Maybe 'notification' and 'action'

met-web/src/components/engagement/view/EngagementView.tsx
```diff
@@ -66,6 +67,27 @@ export const EngagementView = () => {
...
+                                dispatch(
```
jadmsaadaot, line 72: Let's move this to a handle function above so it doesn't bloat the JSX. Also let's not use a hierarchy, data: {header, ..}. jsut header, description, handleClose, handleConfirm, type

met-web/src/store.ts
```diff
@@ -4,13 +4,15 @@ import userSlice from 'services/userService/userSlice';
...
 import notificationSlice from 'services/notificationService/notificationSlice';
+import modalSlice from 'services/modalService/modalSlice';
 
 export const store = configureStore({
     reducer: {
         auth: loginReducer,
         user: userSlice,
         engagement: engagementSlice,
         notification: notificationSlice,
+        modal: modalSlice,
```
jadmsaadaot, line 15: more specific name, notificationModalSlice or better

jadmsaadaot: we might use other custom modals, so let's make sure that this is classified as something specific for notifications or something

met-web/src/components/common/Modals/ConfirmModal.tsx
djnunez-aot: Addressed all comments except changing the data type since not every modal has the same data (the handle confirm is not present in both modals)  thank you for all the feedback

met-web/src/components/common/Modals/ConfirmModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    spacing={1}
+                    sx={{ mt: '1em' }}
+                >
+                    <Stack direction={{ xs: 'column', sm: 'row' }} spacing={1} width="100%" justifyContent="flex-end">
+                        {isSmallScreen ? (
+                            <>
+                                <PrimaryButton onClick={() => confirm()} type="submit" variant={'contained'}>
+                                    Confirm
+                                </PrimaryButton>
+                                <SecondaryButton onClick={handleClose}>Cancel</SecondaryButton>
```
jadmsaadaot, line 52: maybe handleClose here should be handled the same as handleConfirm, if it exists call it and dispatch close modal anyway

met-web/src/components/common/modal.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    return (
+        <Modal
+            sx={{ border: '2px solid red' }}
+            open={open}
+            onClose={handleClose}
+            aria-labelledby="modal-modal-title"
+            aria-describedby="modal-modal-description"
+        >
+            <>
```
jadmsaadaot, line 28: Don't think the fragment are needed here <></>

met-web/src/services/modalService/modalSlice.tsx
```diff
@@ -0,0 +1,27 @@
...
+            state.type = action.payload.type;
+        },
+        closeModal: (state: NotificationModalState) => {
+            state.open = false;
+        },
+    },
+});
+
+// Action creators are generated for each case reducer function
+export const { openModal, closeModal } = notificationModalSlice.actions;
```
jadmsaadaot, line 26: openNotificationModal, closeNotificationModal

met-web/src/components/common/modal.tsx
```diff
@@ -0,0 +1,43 @@
...
+                <ConditionalComponent condition={type === 'update'}>
+                    <UpdateModal header={modalData.header} subText={modalData.subText} handleClose={handleClose} />
+                </ConditionalComponent>
+                <ConditionalComponent condition={type === 'confirm'}>
+                    <ConfirmModal
+                        header={modalData.header}
```
jadmsaadaot, line 34: we can destruct these variables above. const { header, subText, handleConfim, handleClose } = modalData

jadmsaadaot (changes requested):
Minor changes, also please share an example of a dispatch as a comment just so reviewers can see it and suggest if it's too complex or simple enough

met-web/src/components/common/modal.tsx
djnunez-aot: It gives me an error saying MUI modal needs a child component

met-web/src/components/common/modal.tsx
djnunez-aot: will do

djnunez-aot: done good idea

met-web/src/components/common/Modals/ConfirmModal.tsx
djnunez-aot: perfect done

met-web/src/services/modalService/modalSlice.tsx
```diff
@@ -0,0 +1,27 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { NotificationModalState } from './types';
+
+const initialState: NotificationModalState = {
+    open: false,
+    data: { header: '', subText: [] },
+    type: '',
+};
+
+export const notificationModalSlice = createSlice({
```
djnunez-aot, line 10: This is what notification modal dispatch looks like
 dispatch(
                                    openNotificationModal({
                                        open: true,
                                        data: {
                                            header: 'Test',
                                            subText: ['hello'],
                                            handleConfirm: handleStartSurvey,
                                            handleClose: handleClose,
                                        },
                                        type: 'confirm',
                                    }),
                                )
                            }

jadmsaadaot (approved):
Beautiful work. Good job

saravanpa-aot (approved):

PR #577, opened 2022-09-08
Survey Name update

djnunez-aot:
*Issue #404 :*
https://github.com/bcgov/met-public/issues/404

-Icon update

### DIFF
--- a/met-web/src/components/survey/building/index.tsx
+++ b/met-web/src/components/survey/building/index.tsx
@@ -138,6 +138,8 @@ const SurveyFormBuilder = () => {
                     {!isNameFocused ? (
                         <>
                             <MetHeader3
+                                variant="h3"
+                                sx={{ p: 0.5 }}
                                 onClick={() => {
                                     setIsNamedFocused(true);
                                 }}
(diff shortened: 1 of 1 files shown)


met-web/src/components/survey/building/index.tsx
```diff
@@ -138,19 +138,22 @@ const SurveyFormBuilder = () => {
                     {!isNameFocused ? (
                         <>
                             <MetHeader3
+                                variant="h3"
```
tiagograf-AOT, line 141: Is the variant necessary?

djnunez-aot: Marie Anik mentioned she wanted it to be a h3 tag so without the variant in the html code its a p tag

jadmsaadaot: I'm confused here, I though MetHeader3 is an h3 tag according to BC standards. Why would MetHeader3 be a p tag without the variant. Also we created Atomic headers like that to move away from the variant prop no?

jadmsaadaot (changes requested):
Not sure why we are using the variant prop again

met-web/src/components/survey/building/index.tsx
djnunez-aot: The problem was the material ui typography component is defaulted as p tag. If no variant is specified thats how it show up in the html code. No problem I will change the way the Met Headers are implemented in this pr to fix it.

djnunez-aot: It seems this might take a bit more work then expected so Ill remove the variant and fix the header tags in a separate PR.

jadmsaadaot (approved):

VineetBala-AOT (approved):

Issue #341, opened 2022-07-25
create github actions for test deployment


saravanpa-aot:
Place holder to create github actions for missing CI/CD.Details to be filled in later

PR #581, opened 2022-09-09
CD Workflows

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/341

*Description of changes:*
Add continuous deployment workflows for DEV
add deploy workflow for test/prod


jadmsaadaot (approved):

.github/workflows/met-api-cd.yml
```diff
@@ -0,0 +1,68 @@
...
+  APP_NAME: "met-api"
+  TAG_NAME: "dev"
+
+jobs:
+  forms-flow-bpm-cd-by-push:
+    runs-on: ubuntu-20.04
+
+    if: github.event_name == 'push' && github.repository == 'bcgov/met-public'
+    environment:
+      name: "dev"
```
saravanpa-aot, line 33: so do we plan to pass TEST as an arument?

.github/workflows/met-api-cd.yml
```diff
@@ -0,0 +1,68 @@
...
+        shell: bash
+        run: |
+          oc start-build ${{ env.APP_NAME }} --wait
+
+
+      - name: Tag+Deploy for DEV
+        shell: bash
+        run: |
+          oc tag ${{ env.APP_NAME }}:latest ${{ env.APP_NAME }}:${{ env.TAG_NAME }}
+
```
saravanpa-aot, line 64: for TEST or deploy workflows , we dont need to build the image
so we need to have a 
CD-push which can run on code merge
CD which can deploy things(just tag the images)

saravanpa-aot (approved):
Put some comments..
Plz go ahead and try out ur ideas..
I felt like i didnt see a TEST tag in the code ;plz check that out..

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,61 @@
...
+      # - name: Start Build Openshift
+      #   shell: bash
+      #   run: |
+      #     oc start-build met-web --wait
+
+
+      - name: Tag+Deploy
+        shell: bash
+        run: |
+          deploy.sh ${{ env.environment }}
```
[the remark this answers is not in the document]
tiagograf-AOT: since dev is using the 'latest' tag the deployment is triggered always, deployment here would be for test/prod (which we don't have yet, so just test for now!)

.github/workflows/met-api-cd.yml
tiagograf-AOT: This would be the automated github action to deploy to dev, currently our Openshift has the triggers, but we can enable this any time if we decide github is a better place to have control over our CD (when we have a develop branch for example)

.github/workflows/met-api-cd.yml
tiagograf-AOT: I agree for now this will work (when we have a develop branch we will need to build the image based on the 'main' branch before deploying).
This is how it looks now:

- CD-push for each project but not in use yet (the trigger is currently disabled, openshift will build the images on push) when in use this would build and deploy to DEV

- Deploy Action that deploys the entire app (all images get tagged with either test or prod) this can be used to promote dev images into test/prod

.github/workflows/met-api-cd.yml
saravanpa-aot: yup,,we can control DEV, TEST , PROD as an trigger question for deployments..
STAGE can be parameter

Issue #584, opened 2022-09-12
Make all header tags H tags


djnunez-aot:
When the headers were standardized the material UI typography component made all the headers p tags. Need to edit the header components to all be their respective h tags(h1,h2,h3 etc.)

PR #602, opened 2022-09-14
Make Header tags reflect as h tags in html

djnunez-aot:
*Issue #584 :*
https://github.com/bcgov/met-public/issues/584

-Add variants to base component

### DIFF
--- a/met-web/src/components/common/index.tsx
+++ b/met-web/src/components/common/index.tsx
@@ -8,6 +8,7 @@ import {
     Stack,
     IconButton,
 } from '@mui/material';
+import { SxProps } from '@mui/system';
 import { styled } from '@mui/system';
 import EditIcon from '@mui/icons-material/Edit';
 import HighlightOffIcon from '@mui/icons-material/HighlightOff';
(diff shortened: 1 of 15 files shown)


met-web/src/components/common/index.tsx
```diff
@@ -178,30 +178,90 @@ export const modalStyle = {
...
+    onClick?: () => void;
+    color?: string;
+}
+export const MetHeader1 = ({ children, sx, onClick, color, align, style }: HeaderProps) => {
+    return (
+        <Typography
+            sx={{
+                ...sx,
+                fontSize: '2.3rem',
+                fontWeight: 500,
```
saravanpa-aot, line 195: can we define constants for this? I mean for 500

djnunez-aot: done

met-web/src/components/common/index.tsx
```diff
@@ -179,34 +180,96 @@ export const modalStyle = {
...
+export const MetHeader1 = ({ bold, children, sx, onClick, color, align, style }: HeaderProps) => {
+    return (
+        <Typography
+            sx={{
+                ...sx,
+                fontSize: '2.3rem',
+                fontWeight: bold ? 'bold' : MET_Header_Font_Weight,
+                fontFamily: MET_Header_Font_Family,
+            }}
+            style={{ ...style }}
```
jadmsaadaot, line 201: export const SecondaryButton = ({ children, ...rest }: { children: React.ReactNode; [prop: string]: unknown }) => (
    <StyledSecondaryButton {...rest} variant="outlined">
        {children}
    </StyledSecondaryButton>
);

please check the 'rest' prop above and apply it to the Headers. You then don't have to have the onClick, color, align etc. as props. They will all be part of rest. Perhaps maybe the sx keep since you're overriding it and bold since it is not originally in Typography.

jadmsaadaot: Rest literally means whatever other props are passed

jadmsaadaot (changes requested):
use 'rest' prop

jadmsaadaot (approved):

saravanpa-aot (approved):

Issue #589, opened 2022-09-12
Create front end unit test for Engagement Form


tiagograf-AOT:
Make sure 80% of EngagementForm component is covered by unit test so that we can use it as a template for other entities

djnunez-aot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/589?planning-poker) with Zenhub @jadmsaadaot @saravanpa-aot @tiagograf-AOT @VineetBala-AOT

PR #607, opened 2022-09-16
Add unit tests to engagement form page

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/589

*Description of changes:*

- Add mocks for useSelector, useDispatch, useParams, useNavigate, and engagement service methods
- Add 10 unit tests to cover page rendering, warnings, and survey removal


met-web/tests/unit/components/engagementForm.test.tsx
```diff
@@ -1,15 +1,142 @@
...
+
+        await waitFor(() => {
+            expect(screen.getByDisplayValue('Test Engagement')).toBeInTheDocument();
+        });
+
+        const removeSurveyButton = screen.getByTestId('survey-widget/remove');
+
+        fireEvent.click(removeSurveyButton);
+
+        expect(screen.getByText('Do you want to remove this survey?')).toBeInTheDocument();
```
tiagograf-AOT, line 140: It might be worth having an id for things like this too, just bc i see texts are usually fragile, the change often and any change in the wording will cause the test to fail, the same goes for the notification above, I think in general the tests should be agnostic of the contents and more focused on the elements itself

jadmsaadaot: I see your point here, we don't want to get by something that might potentially change. My thoughts here was I wanted to minimize changing the code as much as possible. If I don't have to go to the code and add a test-id then I won't. In this particular case even If I get by id I still need to check the message and make sure it's a warning message. If the modal is no longer a 'warning' modal this test should fail and be updated right?

tiagograf-AOT: That’s right!

met-web/tests/unit/components/engagementForm.test.tsx
```diff
@@ -1,15 +1,142 @@
...
+        useParamsMock.mockReturnValue({ engagementId: 'create' });
+        render(<EngagementForm />);
+
+        const addSurveyButton = screen.getByText('Add Survey');
+
+        fireEvent.click(addSurveyButton);
+
+        expect(openNotificationMock).toHaveBeenNthCalledWith(1, {
+            severity: 'error',
+            text: 'Please save the engagement before adding a survey',
```
tiagograf-AOT, line 123: Like these too, the test shouldn’t care about the message but that there is an error notification and that’s it

tiagograf-AOT, line 123: Like these too, the test shouldn’t care about the message but that there is an error notification and that’s it

met-web/tests/unit/components/engagementForm.test.tsx
jadmsaadaot: getByText is sometimes the closest thing to 'get by semantic' I could use

tiagograf-AOT: Overal it’s looking really good! Good job!

met-web/tests/unit/components/engagementForm.test.tsx
jadmsaadaot: Point noted and awaiting input from @saravanpa-aot as well

jadmsaadaot: Point noted and awaiting input from @saravanpa-aot as well

saravanpa-aot (approved):

tiagograf-AOT (approved):

Issue #558, opened 2022-08-31
Improve and arrive at consistent url patterns in UI


saravanpa-aot:
Engagement related should ideally be 
engagement/:id/action

Survey should ideally be 
survey/:id/action

There should a default route as well /survey/:id should bring a page 

https://met-web-dev.apps.gold.devops.gov.bc.ca/survey/build/29 Should be changed to https://met-web-dev.apps.gold.devops.gov.bc.ca/survey/29/buildd

https://met-web-dev.apps.gold.devops.gov.bc.ca/engagement/form/59 Should be https://met-web-dev.apps.gold.devops.gov.bc.ca/engagement/59/form

@jadmsaadaot @tiago-aot feel free to add your thoughts here

jadmsaadaot:
Also we need to consider action when there is no id, engagement/action, and survey/action.

For the listing pages engagement/listing or list, survey/listing

jadmsaadaot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/558?planning-poker) with Zenhub @VineetBala-AOT @tiagograf-AOT @djnunez-aot

PR #619, opened 2022-09-20
Make url patterns consistent in application

djnunez-aot:
*Issue #558 :*
https://github.com/bcgov/met-public/issues/558

-Update url patterns

### DIFF
--- a/met-web/src/components/engagement/form/AddSurveyBlock.tsx
+++ b/met-web/src/components/engagement/form/AddSurveyBlock.tsx
@@ -96,7 +96,7 @@ export const AddSurveyBlock = () => {
                                 <MetWidget
                                     key={survey.id}
                                     title={survey.name}
-                                    onEditClick={() => navigate(`/survey/build/${survey.id}`)}
+                                    onEditClick={() => navigate(`/survey/:${survey.id}/build`)}
                                     onDeleteClick={() => handleDeleteClick(survey.id, survey.name)}
                                     deleting={isDeletingSurvey}
                                 />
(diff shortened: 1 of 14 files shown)


saravanpa-aot (commented):
Can you put the urls changed in the PR description plz?

met-web/src/components/engagement/form/AddSurveyBlock.tsx
```diff
@@ -96,7 +96,7 @@ export const AddSurveyBlock = () => {
                                 <MetWidget
                                     key={survey.id}
                                     title={survey.name}
-                                    onEditClick={() => navigate(`/survey/build/${survey.id}`)}
+                                    onEditClick={() => navigate(`/survey/:${survey.id}/build`)}
```
jadmsaadaot, line 99: oh I don't think you need to add ':' when you navigate. When you create that route you add it to note that whatever comes after the : is the name of the variable. When you navigate to it you just pass the variable.

met-web/src/routes/UnauthenticatedRoutes.tsx
```diff
@@ -11,7 +11,7 @@ const UnauthenticatedRoutes = () => {
     return (
         <Routes>
             <Route path="/" element={<HomePage />} />
-            <Route path="/engagement/view/:engagementId" element={<EngagementView />} />
+            <Route path="/engagement/:engagementId/view" element={<EngagementView />} />
```
jadmsaadaot, line 14: here you abosulitely have to put the ':' to denote for example the engagementId isn't a path it's a variable name

jadmsaadaot (changes requested):
Minor request to change from singular to plural naming

met-web/src/routes/AuthenticatedRoutes.tsx
```diff
@@ -21,12 +21,12 @@ const AuthenticatedRoutes = () => {
             <Route path="/" element={<LandingPage />} />
             <Route path="/survey/listing" element={<SurveyListing />} />
```
jadmsaadaot, line 22: and let's remove 'listing' and go for /surveys instead of /survey/listing

djnunez-aot: ok perfect will do

Issue #615, opened 2022-09-19
Fix bug in add existing survey to engagement


jadmsaadaot:

jadmsaadaot:
When the select is clicked the page turns white

PR #621, opened 2022-09-20
Add unit tests for engagement listing page

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/615

*Description of changes:*

- Change table linear progress loader implementation by putting it in a table cell and table row instead of a div since div inside table body is invalid html nesting
- Add unit test to check if table is rendering, calling get engagement and displaying the rows
- Add unit test to check the search filter if after adding a search value and clicking search button the get engagements method is called with the search value as a param
- Update engagementForm test. Two methods were fixed and another was removed. That test that was removed tests the modal that appears when remove survey is attempted. The modal now relies on redux and there should be effort to figure out how to mock the redux provider and store to test actions and components that relies on it


met-web/tests/unit/components/EngagementListing.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+const mockSurvey = {
+    ...createDefaultSurvey(),
+    id: 1,
+    name: 'Survey 1',
+    engagement_id: '2',
+};
+
+const mockSurveys = [mockSurvey];
+
+const mockEngagementOne = {
```
saravanpa-aot, line 23: eventually we have to try to move this data to some common place so that it can be reused across different tests...
I am sure react will have patterns for it

saravanpa-aot (approved):

djnunez-aot (approved):

met-web/tests/unit/components/EngagementListing.test.tsx
jadmsaadaot: noted, will do in next test

PR #634, opened 2022-09-21
changes to deploy sentiment analysis

VineetBala-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/600

*Description of changes:*
1) Changes made to add sentiment column to the text area table which will be used to store the sentiment analysis result for each user comment.
2) Added the deployment files for sentiment analysis.


openshift/sentiment_analysis_job_dc.yaml
```diff
@@ -0,0 +1,277 @@
+---
+kind: Template
+apiVersion: v1
+metadata:
+  annotations:
+    description: Deployment template for a payment job.
```
tiagograf-AOT, line 6: Description typo

openshift/sentiment_analysis_job_dc.yaml
```diff
@@ -0,0 +1,277 @@
...
+  displayName: Database type to use
+  description: Database type to use. SQL_SERVER or POSTGRES
+  required: true
+  value: 
+- name: MODEL_ID
+  displayName: Model ID for sentiment analysis
+  description: Model ID for sentiment analysis
+  required: true
+  value: 
+- name: FORMSFLOW_CONFIG_NAME
```
tiagograf-AOT, line 273: Let’s change the variable name or drop it assuming we don’t need formsflow

met-cron/migrations/versions/e817bbaef67e_add_sentiment_to_textarea.py
```diff
@@ -0,0 +1,26 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.add_column('response_type_textarea', sa.Column('sentiment', sa.String(length=50), nullable=True))
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    pass
```
tiagograf-AOT, line 25: When downgrading he version ideally we want to revert the changes made, in this case drop the column

openshift/sentiment_analysis_job_bc.yaml
```diff
@@ -0,0 +1,105 @@
...
+  displayName: Docker File Path
+  description: The path to the docker file defining the build.
+  required: false
+  value: 
+- name: DOCKER_CRED_NAME
+  displayName: Docker cred secret name
+  description: Docker cred secret name
+  required: true
+  value: 
+- name: FORMIO_SOURCE_REPO_URL
```
tiagograf-AOT, line 91: This goes with e same idea of the other variable, we can rename it or drop it

PR #647, opened 2022-09-22
Add unit tests for survey listing

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/646

*Description of changes:*

- Add unit test to check table rendering and rows fetching
- Add unit test to check search filter working and calling proper service method

### DIFF
--- a/met-web/src/components/survey/submit/SurveySubmitWrapped.tsx
+++ b/met-web/src/components/survey/submit/SurveySubmitWrapped.tsx
@@ -30,17 +30,17 @@ const SurveySubmitWrapped = () => {
                 </Grid>
                 <Grid item xs={12}>
                     <MetPaper elevation={2}>
-                        <ConditionalComponent condition={isTokenValid && savedSurvey.engagement}>
+                        <ConditionalComponent condition={isTokenValid && Boolean(savedSurvey.engagement)}>
                             <SurveyForm
                                 handleClose={() => {
-                                    navigate(`/engagements/${savedSurvey.engagement.id}/view`);
+                                    navigate(`/engagements/${savedSurvey.engagement?.id}/view`);
                                 }}
                             />
                         </ConditionalComponent>
                         <InvalidTokenModal
-                            open={!isTokenValid && savedSurvey.engagement}
+                            open={!isTokenValid && Boolean(savedSurvey.engagement)}
                             handleClose={() => {
-                                navigate(`/engagements/${savedSurvey.engagement.id}/view`);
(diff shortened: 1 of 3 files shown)


met-web/src/components/survey/submit/SurveySubmitWrapped.tsx
```diff
@@ -30,17 +30,17 @@ const SurveySubmitWrapped = () => {
...
                             <SurveyForm
                                 handleClose={() => {
-                                    navigate(`/engagements/${savedSurvey.engagement.id}/view`);
+                                    navigate(`/engagements/${savedSurvey.engagement?.id}/view`);
                                 }}
                             />
                         </ConditionalComponent>
                         <InvalidTokenModal
-                            open={!isTokenValid && savedSurvey.engagement}
+                            open={!isTokenValid && Boolean(savedSurvey.engagement)}
```
saravanpa-aot, line 41: what does the boolean do here?

met-web/tests/unit/components/surveyListing.test.tsx
```diff
@@ -1,20 +1,145 @@
...
+        await waitFor(() => {
+            expect(screen.getByText('Survey One')).toBeInTheDocument();
+            expect(screen.getByText('2022-09-14')).toBeInTheDocument();
+            expect(screen.getByText('Draft')).toBeInTheDocument();
+
+            expect(screen.getByText('Survey Two')).toBeInTheDocument();
+            expect(screen.getByText('2022-09-15')).toBeInTheDocument();
+            expect(screen.getByText('Published')).toBeInTheDocument();
+            expect(screen.getByText('2022-09-19')).toBeInTheDocument();
+            expect(screen.getByText('View Report')).toBeInTheDocument();
```
saravanpa-aot, line 110: Love this one..

saravanpa-aot (approved):

tiagograf-AOT (approved):

met-web/src/components/survey/submit/SurveySubmitWrapped.tsx
jadmsaadaot: This change really isn't part of my task. I had a code strictness error in the Web build action and it complained that savedSurvey.engagement could be null or undefined and there's a && before it meaning it is expecting it to be Boolean. So when you wrap Boolean() around any object it really checks if it's truthy or Falsey. and spits out a boolean So null, undefined will be false and anything else true. With this the error went away

PR #654, opened 2022-09-28
Update Postman collections

djnunez-aot:
-Restructure Postman files

image


saravanpa-aot (dismissed):
I think MET as one collection will be ideal.
under that we can have several requests for each thing

jadmsaadaot (approved):

Issue #510, opened 2022-08-24
update of an engagement should use Patch instead of PUT


saravanpa-aot:
![putvspatch.png](https://images.zenhubusercontent.com/61670716af0598935f770d72/b736539e-bf86-46ca-a4b1-2637e88a61cc)update of engagement as name , published date etc are handled in PUT where the whole resource is updated which is not ideal.
PATCH is recommended in this case and UI has to send only whats getting updated.

jadmsaadaot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/510?planning-poker) with Zenhub @djnunez-aot @saravanpa-aot @tiagograf-AOT @VineetBala-AOT

PR #655, opened 2022-09-28
Replace PUT command with PATCH for updating engagement

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/510

*Description of changes:*

- validate engagement form data with the schema
- dynamically update fields in the back-end
- partially load data in engagement schema
- Only send changed data in front for update

### DIFF
--- a/met-api/src/met_api/models/engagement.py
+++ b/met-api/src/met_api/models/engagement.py
@@ -5,7 +5,7 @@ Manages the engagement
 
 from __future__ import annotations
 from datetime import datetime
-from typing import List
+from typing import List, Optional
 from sqlalchemy import asc, desc
 from sqlalchemy.dialects.postgresql import JSON
 from sqlalchemy.sql import text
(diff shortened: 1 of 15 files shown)


met-api/src/met_api/models/engagement.py
```diff
@@ -141,6 +141,19 @@ def update_engagement(cls, engagement: EngagementSchema) -> DefaultMethodResult:
         db.session.commit()
         return DefaultMethodResult(True, 'Engagement Updated', engagement_id)
 
+    @classmethod
+    def edit_engagement(cls, engagement_data: dict) -> DefaultMethodResult:
```
saravanpa-aot, line 145: Just a quick review..
Let model return the model .We are trying to move away from DefaultMethodResult..We will remove it eventually..
If not record: can return a None
If its an error , it shud throw an error

jadmsaadaot: Got it, will do

tiagograf-AOT (approved):

VineetBala-AOT (approved):

PR #705, opened 2022-10-06
Add unit test for feedback listing

djnunez-aot:
*Issue #699 :*
https://github.com/bcgov/met-public/issues/699

-Add feedback listing unit test

### DIFF
--- a/met-web/tests/unit/components/feedbackListing.test.tsx
+++ b/met-web/tests/unit/components/feedbackListing.test.tsx
@@ -5,8 +5,7 @@ import { setupEnv } from './setEnvVars';
 import * as reactRedux from 'react-redux';
 import * as feedbackService from 'services/feedbackService/index';
 import * as notificationSlice from 'services/notificationService/notificationSlice';
-import { createDefaultFeedback } from 'models/feedback';
-import { CommentTypeEnum, SourceTypeEnum } from 'models/feedback';
+import { createDefaultFeedback, CommentTypeEnum, SourceTypeEnum } from 'models/feedback';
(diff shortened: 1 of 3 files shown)


tiagograf-AOT (approved):

saravanpa-aot (commented):
can we have a test for no feedback as well? when there is no data , how the table behaves?
Also any way to check the paginator in the tests?is it easy to validate total=2 in the below test

djnunez-aot:
Will set that up now

tiagograf-AOT (approved):

met-web/tests/unit/components/feedbackListing.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    mockFeedbackOne,
+                    mockFeedbackOne,
+                    mockFeedbackOne,
+                    mockFeedbackOne,
+                    mockFeedbackOne,
+                    mockFeedbackOne,
+                    mockFeedbackOne,
+                    mockFeedbackOne,
+                    mockFeedbackOne,
+                ],
```
saravanpa-aot, line 125: can u do a for loop or  Array.from({length:11} syntax or array.fill to do the same thing rather than writing this 11 times? i think there will be  lot of different ways to achieve this..

saravanpa-aot (dismissed):

met-web/tests/unit/components/feedbackListing.test.tsx
djnunez-aot: Thank you I never thought about that. Cleaned that up!

jadmsaadaot (approved):

Issue #98, opened 2022-05-10
As an Admin, I can select an existing survey to clone it


sidtobias:
**As an** Admin

**I want** to select an existing survey and clone it

**so that** I can use it as a template that I can edit to customize

### Scenario 1: 
Admin clicks the "Create Survey" button on the Survey listing page OR from the engagement page.

Admin sees the "Create Survey" screen. 

Section Header: "Create Survey" (+ edit icon = widget name is editable)
Sees available options depending on the page they are coming from (https://app.zenhub.com/workspaces/met-ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/97 OR https://app.zenhub.com/workspaces/met-ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/47)
"Save & Continue" and "Cancel" buttons

Admin selects the "Clone an existing Survey" and sees:
Drop-down: Select Engagement (optional)
Drop-down:  Select Survey 
Input field to "Enter Survey Name" (pre-populated with survey name)
"Save & Continue" and "Cancel" buttons

### Scenario 2:
User clicks "Add Survey ", goes to the survey panel with all the questions from the cloned/existing survey. 

### Scenario 3:
Matches wireframes
(from Survey listing page)
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/01a280d9-bc72-4fb3-96ed-1e78bcdac199
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/719f3a59-f300-4bb3-a291-98cafc176a17
(from Engagement page)
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/bf0d9afb-1c4b-4ab3-a08f-8ed5a7b35163/
https://xd.adobe.com/view/e38dc788-12d5-4844-87ca-149d1d32d395-b6de/screen/c180802e-cc2d-4bde-9ae6-231dcce70e65/

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

jadmsaadaot:
[...]

PR #727, opened 2022-10-14
Clone survey

djnunez-aot:
*Issue #98 :*
https://github.com/bcgov/met-public/issues/98

-Add Cloning options form

-Update context to fetch engagements


met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,232 @@
...
+            const fetchedEngagements = await getEngagements({
+                page: page,
+                size: size,
+                sort_order: sort_order,
+                search_text: search_text,
+            });
+            console.log('FETCH!');
+            console.log(fetchedEngagements);
+            setAvailableEngagements(
+                fetchedEngagements.items.filter((engagement, index) => engagement.surveys.length !== 0),
```
jadmsaadaot, line 87: you didn't use/need index here

met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,232 @@
...
+    };
+
+    return (
+        <Grid container direction="row" alignItems="flex-start" justifyContent="flex-start" item xs={12} spacing={2}>
+            <Grid item xs={6}>
+                <MetLabel sx={{ marginBottom: '2px', display: 'flex' }}>
+                    Select Engagement <Typography sx={{ ml: 1, color: '#ACA9A9' }}>(optional)</Typography>
+                </MetLabel>
+
+                <Autocomplete
```
jadmsaadaot, line 147: select engagement is optional, should be used to filter the surveys select when used. If surveys are 100 and select engagement is used to select an engagement then of those 100 we return only the surveys that have engagement_id equal to the selectedEngagement id in our case it will always be one survey until later when we might have multiple surveys for one engagement, or for one project.

met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,232 @@
...
+                        setSelectedEngagement(engagement);
+                        if (engagement !== null) setSelectedSurvey(engagement.surveys[0]);
+                    }}
+                    disabled={loadingEngagements}
+                />
+            </Grid>
+            <Grid item xs={12}>
+                <MetLabel sx={{ marginBottom: '2px' }}>Select Survey</MetLabel>
+                <Autocomplete
+                    id="survey-selector"
```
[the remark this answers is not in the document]
djnunez-aot: awesome

met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+    useEffect(() => {
+        if (!availableEngagements) {
+            handleFetchEngagements(1, 10, 'asc', '');
+        }
+    }, [availableEngagements]);
+
+    const handleSave = async () => {
+        if (!selectedSurvey) {
+            if (selectedEngagement) return;
```
jadmsaadaot, line 111: always have to select a survey. if(!selectedSurvey) dispatch error or show error text for selectSurvey always

met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,232 @@
...
+        } catch (error) {
+            dispatch(
+                openNotification({ severity: 'error', text: 'Error occurred while fetching available engagements' }),
+            );
+        }
+    };
+
+    useEffect(() => {
+        if (!availableSurveys) {
+            handleFetchSurveys(1, 10, 'asc', '');
```
jadmsaadaot, line 99: we're always only calling for 10? what if there's more?

met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,232 @@
...
+        }
+    }, [availableSurveys]);
+
+    useEffect(() => {
+        if (!availableEngagements) {
+            handleFetchEngagements(1, 10, 'asc', '');
```
jadmsaadaot, line 105: what if there's more than 10?

met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,232 @@
...
+                    renderInput={(params) => (
+                        <TextField
+                            {...params}
+                            label=" "
+                            value={selectedEngagement && selectedSurvey != null ? selectedSurvey.name : ''}
+                            InputLabelProps={{
+                                shrink: false,
+                            }}
+                            onChange={(e) => {
+                                handleFetchSurveys(1, 10, 'asc', e.target.value);
```
jadmsaadaot, line 185: best to  debounce or throttle this since on each keystroke an API call will happen

met-web/src/components/survey/create/CloneOptions.tsx
jadmsaadaot: Perhaps for now let it fetch all surveys or atleast specify a big number (2000) page size. Until we implement proper pagination and lazy loading for it. @saravanpa-aot thoughts?

jadmsaadaot (changes requested):
Select survey should be mandatory, select engagement should be treated as a filter of select surveys. A way to narrow user's options. and A few other suggestions

met-web/src/components/survey/create/CloneOptions.tsx
djnunez-aot: ill set it up for fetching 2000 i was trying to make it so that we fetch 10 if user cant find it they type in the search bar and we fetch what we need

met-web/src/components/survey/create/CloneOptions.tsx
djnunez-aot: i think thats a better idea

met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,232 @@
...
+                    renderInput={(params) => (
+                        <TextField
+                            {...params}
+                            label=" "
+                            InputLabelProps={{
+                                shrink: false,
+                            }}
+                            fullWidth
+                            onChange={(e) => {
+                                handleFetchEngagements(1, 10, 'asc', e.target.value);
```
saravanpa-aot, line 159: 1,10 i think are the page size rite? can you please use a constant here?
https://refactoring.guru/replace-magic-number-with-symbolic-constant

saravanpa-aot (approved):

met-web/src/components/survey/create/CloneOptions.tsx
jadmsaadaot: Yup, figured you were going for that. Great idea. I'm thinking we can extend it and implement something like also when user scrolls and reaches end of list another page of items will be loaded. We just need to figure how to do that, seems doable and all, just feels a bit tricky

met-web/src/components/survey/create/CloneOptions.tsx
djnunez-aot: since some of the engagements share the same name there's a repeated key error without the index parameter

met-web/src/components/survey/create/CloneOptions.tsx
djnunez-aot: smart

met-web/src/components/survey/create/CloneOptions.tsx
djnunez-aot: resolved

tiagograf-AOT (approved):

met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,225 @@
...
+            <Grid item xs={12}>
+                <MetLabel sx={{ marginBottom: '2px' }}>Select Survey</MetLabel>
+                <Autocomplete
+                    id="survey-selector"
+                    options={selectedEngagement ? selectedEngagement.surveys : availableSurveys || []}
+                    renderInput={(params) => (
+                        <TextField
+                            {...params}
+                            label=" "
+                            value={selectedEngagement && selectedSurvey != null ? selectedSurvey.name : ''}
```
jadmsaadaot, line 176: selectedSurvey?.name || '' maybe?

met-web/src/components/survey/create/CloneOptions.tsx
```diff
@@ -0,0 +1,225 @@
...
+                            InputLabelProps={{
+                                shrink: false,
+                            }}
+                            fullWidth
+                        />
+                    )}
+                    getOptionLabel={(survey: Survey) =>
+                        selectedEngagement && selectedSurvey ? selectedSurvey.name : survey.name
```
jadmsaadaot, line 184: Should this not be always option: Survey => option.name. What is the logic behind this ternary?

met-web/src/components/survey/create/CloneOptions.tsx
djnunez-aot: you are right these are not needed

met-web/src/components/survey/create/CloneOptions.tsx
djnunez-aot: it was from an earlier iteration good catch

Issue #396, opened 2022-08-08
Allow rollback when error occurs between database transactions


tiagograf-AOT:
If an error occur during multi step database statements the system should be able to rollback the changes previously made.

jadmsaadaot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/396?planning-poker) with Zenhub @tiagograf-AOT @saravanpa-aot @VineetBala-AOT

PR #728, opened 2022-10-14
Add DB transaction support

tiagograf-AOT:
Issue #:
https://github.com/bcgov/met-public/issues/396

Description of changes:
Add support for scoped db sessions
Change submission to use scoped session

### DIFF
--- a/met-api/src/met_api/models/comment.py
+++ b/met-api/src/met_api/models/comment.py
@@ -101,7 +101,7 @@ class Comment(db.Model):
         )
 
     @classmethod
-    def add_all_comments(cls, comments: list, session = None) -> DefaultMethodResult:
+    def add_all_comments(cls, comments: list, session=None) -> DefaultMethodResult:
         """Save comments."""
         new_comments = [cls.__create_new_comment_entity(comment) for comment in comments]
         if session is None:
--- a/met-api/src/met_api/models/comment.py
+++ b/met-api/src/met_api/models/comment.py
@@ -101,11 +101,14 @@ class Comment(db.Model):
         )
 
     @classmethod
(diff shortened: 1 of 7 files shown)


met-api/src/met_api/models/comment.py
```diff
@@ -101,11 +101,14 @@ def __create_new_comment_entity(comment):
...
+    def add_all_comments(cls, comments: list, session=None) -> DefaultMethodResult:
         """Save comments."""
         new_comments = [cls.__create_new_comment_entity(comment) for comment in comments]
-        db.session.add_all(new_comments)
-        db.session.commit()
+        if session is None:
+            db.session.add_all(new_comments)
+            db.session.commit()
+        else:
+            session.add_all(new_comments)
```
saravanpa-aot, line 111:
```
    if session is None:
            db.session.add_all(new_comments)
            db.session.commit()
        else:
            session.add_all(new_comments)
  ```
  to 
  ```
      db.session.add_all(new_comments)
      if session is None:
            db.session.commit()
            ```          
Does this work? the intention is that , if there is a session , dont commit rite? so this simplification might work

saravanpa-aot (approved):
Loved the idea of making into contextmanager..

met-api/src/met_api/models/comment.py
tiagograf-AOT: It works as today because the session object is the same as the db.session, if we decide to create a new scoped session in the future I think that could cause a problem

saravanpa-aot: Makes sense..

jadmsaadaot (approved):

VineetBala-AOT (approved):

Issue #740, opened 2022-10-18
As an MET system user, I want to see and approve/reject all the written comments made by one person on one page.


marieanik:
As an MET system user, I want to see and approve/reject all the written comments made by one person on one page.

Scenario 1:
MET survey has more than one single or multiple line answer question. **On the comments review page, User sees all the comments, separated by a #f2f2f2 line**. Comments (all) can be approved or rejected. 

![Screen Shot 2022-11-08 at 1.16.59 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/231eba79-6ea0-4eb7-94d3-48cd10bc4b35)
https://xd.adobe.com/view/7d5f6547-2523-484d-8101-a7587f9076c8-280c/screen/8cc0f2d1-ba05-48fd-9b6e-ec7b321e8c47

Consideration: leave space for sentiment analysis and VC icons on the right of the comments. Reason for rejection will be added later.

![Screen Shot 2022-10-18 at 11.48.32 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/ae9b56ae-ae96-4f28-b67f-0ceae285d4a2)

carlygreenehill:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/740?planning-poker) with Zenhub @djnunez-aot @jadmsaadaot @saravanpa-aot @tiagograf-AOT @VineetBala-AOT

PR #808, opened 2022-11-01
Review multiple comments

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/740

*Description of changes:*
Changes to review comments to allow for mutiple comment review


met-api/migrations/versions/f2b3f08c8d60_alter_comment_add_submission_id.py
```diff
@@ -0,0 +1,52 @@
...
+revision = 'f2b3f08c8d60'
+down_revision = 'c19bc1af9f2b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    op.add_column('comment', sa.Column('submission_id', sa.Integer(), nullable=True))
+    op.add_column('comment', sa.Column('component_id', sa.String(), nullable=True))
```
saravanpa-aot, line 22: Should we call component_id as question_id ? was wondering how to make it easy to convey that component_id is referring to the question thats being asked?

met-api/src/met_api/models/comment.py
```diff
@@ -112,12 +129,12 @@ def add_all_comments(cls, comments: list, session=None) -> DefaultMethodResult:
         return DefaultMethodResult(True, 'Comments Added', 1)
 
     @classmethod
-    def update_comment_status(cls, comment_id, status_id, reviewed_by) -> DefaultMethodResult:
+    def update_comment_status(cls, submission_id, status_id, reviewed_by) -> DefaultMethodResult:
```
saravanpa-aot, line 132: I think this is an existing code , but feel free to return CommentModel | None and remove the defaultmethodresult

met-api/src/met_api/models/submission.py
```diff
@@ -52,6 +52,8 @@ def create(cls, submission: SubmissionSchema, session=None) -> DefaultMethodResu
             db.session.commit()
         else:
             session.add(new_submission)
+            session.flush()
+            session.refresh(new_submission)
```
saravanpa-aot, line 56: I never had to refresh explicitly , whats the reason here?

met-api/src/met_api/schemas/comment.py
```diff
@@ -24,5 +24,22 @@ class Meta:  # pylint: disable=too-few-public-methods
...
+            return None
+        component_label = [
+            component.get(
+                'label',
+                None) for component in components if component.get(
+                'key',
+                None) == obj.component_id]
+        if len(component_label) == 0:
+            return None
+        return component_label[0]
```
saravanpa-aot, line 45: i am just thinking aloud , does it help to store the question as such in comments table? i know its duplication and eats up space , but added value is performance and we can remove this logic...what do you think?
Its not ideal i know , just hate the hassle of going through the compoent json to extract the question

met-web/src/services/commentService/index.tsx
```diff
@@ -40,6 +40,22 @@ export const getCommentsPage = async ({
     );
 };
 
+export const getSubmissionComments = async (submissionId: number): Promise<Comment[]> => {
+    const url = replaceUrl(Endpoints.Comment.GET_SUBMISSION, 'submission_id', String(submissionId));
+    if (!submissionId || isNaN(Number(submissionId))) {
+        return Promise.reject('Invalid Submission Id ' + submissionId);
+    }
```
saravanpa-aot, line 47: remove this plz if you think there is no additional value , i think it gets carried over from the previous code...

met-web/src/components/comments/admin/review/CommentReview.tsx
```diff
@@ -47,24 +54,24 @@ const CommentReview = () => {
...
+            navigate(`/surveys/${comments[0].survey_id}/comments`);
         } catch (error) {
-            dispatch(openNotification({ severity: 'error', text: 'Error occurred while sending comment review.' }));
+            dispatch(openNotification({ severity: 'error', text: 'Error occurred while sending comments review.' }));
             setIsSaving(false);
         }
     };
 
-    const { id, comment_status, reviewed_by, submission_date, review_date, text } = comment;
+    const { submission_id, comment_status, reviewed_by, submission_date, review_date } = comments[0];
```
saravanpa-aot, line 67: not exactly part of this task 
but may be we should write some comment about why we deal with comments[0] alone ;
explain the logic that comments are approved togother and 0 and n will have same verdict..

saravanpa-aot (approved):
Well Done!!!

met-api/src/met_api/schemas/comment.py
tiagograf-AOT: Yep I thought about that, also there's the versioning of the question (lets say someone changes the question text in the future) this would keep exctly what was asked at the time of the comment creation, that said the only reason I stored the id instead of the question is bc of duplication, but if we think its enough value i'm happy to store it too

met-api/src/met_api/models/submission.py
tiagograf-AOT: because its part of the transaction for some reason the id does not get reflected in the model object until a flush and refresh, that said i haven't tested with the flush only

met-api/migrations/versions/f2b3f08c8d60_alter_comment_add_submission_id.py
tiagograf-AOT: since we are going through the list of components to fetch its ids that what I thought was the name of it, i don't think. 'question' works too because it could be labeled as something like 'Leave your comment' which technically is not a question either

tiagograf-AOT: does survey_component_id sounds any better?

met-api/src/met_api/models/submission.py
tiagograf-AOT: refresh removed since flush is enough to return its id!

met-web/src/services/commentService/index.tsx
tiagograf-AOT: removed

met-web/src/components/comments/admin/review/CommentReview.tsx
tiagograf-AOT: comment added

met-api/src/met_api/models/submission.py
saravanpa-aot: yeah..for id , flush does the job

met-api/tests/utilities/factory_scenarios.py
```diff
@@ -163,7 +163,7 @@ class TestContactInfo(dict, Enum):
 
     contact1 = {
         'name': fake.name(),
-        'title': fake.job(max_length=50),
+        'title': fake.job(),
```
jadmsaadaot, line 166: is this an intended change? fake.job sometimes springs up a very long title and title is title = db.Column(db.String(50))

jadmsaadaot (approved):

met-api/tests/utilities/factory_scenarios.py
tiagograf-AOT: This was breaking the tests earlier with a message saying max_length arg was not supported, i'm not sure if there was a version change that broke it but removing it fixed the problem!

PR #813, opened 2022-11-04
Add Value Component options to form builder

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/801

*Description of changes:*

- Create custom checkbox and textarea components in met-formio
- update met-formio version
- add the vc options in the formbuilder options under basic

### DIFF
--- a/met-web/src/App.tsx
+++ b/met-web/src/App.tsx
@@ -13,6 +13,7 @@ import { Notification } from 'components/common/notification';
 import PageViewTracker from 'routes/PageViewTracker';
 import { NotificationModal } from 'components/common/modal';
 import { FeedbackModal } from 'components/common/Modals/Feedback';
+import { AppConfig } from 'config';
 
 const App = () => {
     const drawerWidth = 240;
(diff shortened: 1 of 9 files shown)


met-web/src/App.tsx
```diff
@@ -25,6 +26,8 @@ const App = () => {
         UserService.initKeycloak(dispatch);
     }, [dispatch]);
 
+    localStorage.setItem('apiurl', String(AppConfig.apiUrl));
```
saravanpa-aot, line 29: hm..General principle is to stick to sessionstorage unless there is a motive to store items in localstorage ;like track a user even if he closes browser..I guess here good enough to be in sessionstorage rite?

saravanpa-aot (approved):
We have to discuss with Tiago on the noEval 
Also plz have a look at the localstorage comment

met-web/src/App.tsx
jadmsaadaot: Done, replaced by sessionStorage

VineetBala-AOT (approved):

Issue #784, opened 2022-10-25
As a public user, I'd like to see the Who is Listening Widget on Engagement Page


tiagograf-AOT:
As an Engagement specialist I want to be able to see the added who is listening widget in the engagement preview page so that i can confirm that the information is correct

this user story also applies to a public user:

As a BC Citizen I want to be able to see who is listening to an engagement so that I feel heard and more connected to issue/cause raised by the engagement.

carlygreenehill:
See cards 142, 143, 144 for similar cards.

carlygreenehill:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/784?planning-poker) with Zenhub @djnunez-aot @jadmsaadaot @saravanpa-aot @tiagograf-AOT @VineetBala-AOT

PR #816, opened 2022-11-05
Updates to who is listening widget

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/784

*Description of changes:*
Update who is listening widget to use contact information stored in the engagement.

### DIFF
--- a/met-web/src/components/engagement/view/ActionContext.tsx
+++ b/met-web/src/components/engagement/view/ActionContext.tsx
@@ -4,6 +4,9 @@ import { getEngagement, patchEngagement } from '../../../services/engagementServ
 import { createDefaultEngagement, Engagement } from '../../../models/engagement';
 import { useAppDispatch } from 'hooks';
 import { openNotification } from 'services/notificationService/notificationSlice';
+import { getWidgets } from 'services/widgetService';
+import { WhoIsListeningWidget, WidgetType } from 'models/widget';
+import { getContact } from 'services/contactService';
 
 interface EngagementSchedule {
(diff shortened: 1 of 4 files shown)


met-web/src/components/engagement/view/ActionContext.tsx
```diff
@@ -72,7 +77,36 @@ export const ActionProvider = ({ children }: { children: JSX.Element | JSX.Eleme
                 );
             }
         };
+
+        const fetchWidgets = async () => {
+            try {
+                const result = await getWidgets(Number(engagementId));
+                const whoIsListeningWidget = result.find((w) => w.widget_type_id === WidgetType.WhoIsListening);
+                if (!whoIsListeningWidget) return;
```
jadmsaadaot, line 85: All good for now. But when we have several widget types this method will become bloated. Perhaps we should remain general at this point like just get widgets (no contacts are any other data). We can create a generic widget card which checks for the type and renders the corresponding widget. If type is who is listening render who is listening widget (switch case maybe). Then that who is listening widget fetches its data.

jadmsaadaot: There's also the issue of sorting. At some point we'll have a custom sorting for the widgets which could be easier if we just call get widgets sorted then render them agnostically before each widget calls its data.

jadmsaadaot (approved):
Approving to merge it and address comment in another PR

saravanpa-aot (approved):

PR #817, opened 2022-11-05
Change review comments table to group comments by submission

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/740

*Description of changes:*
Changnes review comment table to group records by submission

### DIFF
--- a/met-api/src/met_api/models/submission.py
+++ b/met-api/src/met_api/models/submission.py
@@ -5,9 +5,13 @@ Manages the Submission
 from datetime import datetime
 from typing import List
 
-from sqlalchemy import ForeignKey
+from sqlalchemy import TEXT, ForeignKey, asc, cast, desc, text
 from sqlalchemy.dialects import postgresql
+from sqlalchemy.orm import aliased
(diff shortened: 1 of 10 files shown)


jadmsaadaot (commented):
Just a thought. Since now the comments listing is actually submission listing we might need to change the name in the UI to submission listing or something like that. It might be a bit confusing for a developer to see a comments listing page getting submissions

met-api/src/met_api/models/submission.py
```diff
@@ -81,3 +86,39 @@ def get_engaged_users(cls, engagement_id) -> List[User]:
             .filter(Survey.engagement_id == engagement_id)\
             .all()
         return users
+
+    @classmethod
+    def get_by_survey_id_paginated(cls, survey_id, pagination_options: PaginationOptions, search_text=''):
```
jadmsaadaot, line 91: Just thinking out loud. What made us need to query the submission table for this purpose. Why not query the comments table and  group by submission id?

Issue #785, opened 2022-10-25
As an admin, I'd like to add a photo to the contact in Who's Listening widget


tiagograf-AOT:
Need to display an icon if no photo is added. *will be provided soon (working with Erick Ortiz, EOA Graphic Designer to get it). 
Image size is 125 px x 125px, not resizeable.
What are the preferred image formats? .jpg or .png

PR #846, opened 2022-11-14
Add upload image to contact

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/785

*Description of changes:*

- Add image upload functionality for contact Avatar
- Display saved image in public page
- Fix sorting of widget items


djnunez-aot (approved):

met-api/migrations/versions/03ee1815f6a6_.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '03ee1815f6a6'
+down_revision = '9536f547cdd5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('contact', sa.Column('avatar_filename', sa.String(), nullable=True))
```
saravanpa-aot, line 21: are we planning to just store the name or the full url?

met-api/src/met_api/services/contact_service.py
```diff
@@ -11,13 +12,20 @@ class ContactService:
     def get_contact_by_id(contact_id):
         """Get contact by id."""
         contact_record = Contact.get_contact_by_id(contact_id)
-        return ContactSchema().dump(contact_record)
+        contact = ContactSchema().dump(contact_record)
+        contact['avatar_url'] = ObjectStorageService.get_url(contact.get('avatar_filename', None))
```
saravanpa-aot, line 16: i forgot whats in the O.get_url() does ;just thinking will it be faster if we store the full url in the DB and avoid this logic? We can discuss this and weigh each approach

jadmsaadaot: We can discuss this. The reasoning behind it is that this will work even if host changed.

saravanpa-aot (approved):

met-api/migrations/versions/03ee1815f6a6_.py
jadmsaadaot: As agreed in our chat. We will discuss this with @tiago-aot in a meeting

Issue #860, opened 2022-11-16
Schedule Engagement banner shows wrong time


tiagograf-AOT:
Engagement was scheduled at 4:52 PM but the banner displays time at 00:52
![Screenshot 2022-11-15 at 4.53.06 PM.png](https://images.zenhubusercontent.com/6254616e054e9b1da214d3b1/410ec59f-4ca3-45ce-9514-1124f068768e)

VineetBala-AOT:
This is looking good. Tested in a different time zone as well and the time showing up is PST.

![image.png](https://images.zenhubusercontent.com/614119663ff89097e85e7bb8/5ca5cf1f-c95f-4e32-8fc6-766bbe153e03)

![image.png](https://images.zenhubusercontent.com/614119663ff89097e85e7bb8/45ca385b-926d-43a5-aea7-9a5aabc83c20)

carlygreenehill:
Looking good to me (I work in PST)

PR #863, opened 2022-11-16
Change standard timezone to PST

djnunez-aot:
*Issue #860:*
https://github.com/bcgov/met-public/issues/860

*Description of changes:*

### DIFF
--- a/met-web/src/components/common/dateHelper.tsx
+++ b/met-web/src/components/common/dateHelper.tsx
@@ -11,7 +11,7 @@ const formatInTimeZone = (date: string, fmt: string, tz: string) =>
 
 export const formatDate = (d: string, formatString = 'yyyy-MM-dd') => {
     if (d) {
-        return formatInTimeZone(d, formatString, 'UTC');
+        return formatInTimeZone(d, formatString, 'PST');
     } else {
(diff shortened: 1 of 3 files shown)


tiagograf-AOT:
This will impact anywhere the format function is used, from what I noticed this problem only happens with the scheduling date/time, can you check if there’s any impact in the rest of the app?

met-web/src/components/common/dateHelper.tsx
```diff
@@ -11,7 +11,7 @@ const formatInTimeZone = (date: string, fmt: string, tz: string) =>
 
 export const formatDate = (d: string, formatString = 'yyyy-MM-dd') => {
     if (d) {
```
jadmsaadaot, line 13: If along the way we can change this variable name from d to date it would be great.

jadmsaadaot:
Test failed since the format is outputting a date more accurate to us while test still expects "2022-09-17"

saravanpa-aot (changes requested):
Is there a way to handle without specifying time zone?
Normally moment will have a way to convert to local timezone instead of specifying PST.. it shud ideally dependent on users time zone i guess

PR #890, opened 2022-11-22
Added sort_index for widgets

saravanpa-aot:
*Issue #:*

*Description of changes:*

- Added sort_index column
- modified logic to increase sort_index while creating a new widget
- Added new endpoint for sorting the widgets

### DIFF
--- a/met-api/migrations/versions/e37d79be3a05_added_sort_for_widgets.py
+++ b/met-api/migrations/versions/e37d79be3a05_added_sort_for_widgets.py
@@ -1,0 +1,28 @@
+"""added sort for widgets
+
+Revision ID: e37d79be3a05
+Revises: 9f86fdcfb248
+Create Date: 2022-11-20 17:25:42.697782
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'e37d79be3a05'
+down_revision = '9f86fdcfb248'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### Add Sort index to widget.###
+    op.add_column('widget', sa.Column('sort_index', sa.Integer(), nullable=False, server_default='1'))
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### Remove the sort index column. ###
+    op.drop_column('widget', 'sort_index')
+    # ### end Alembic commands ###
--- a/met-api/src/met_api/services/widget_service.py
+++ b/met-api/src/met_api/services/widget_service.py
@@ -53,7 +53,7 @@ class WidgetService:
 
     @staticmethod
     def sort_widget(engagement_id, widget_items: list, user_id=None):
-        """Sort widget items."""
+        """Sort widgets."""
         WidgetService._validate_widget_ids(engagement_id, widget_items)
 
         widget_sort_mappings = [{
--- a/met-api/src/met_api/services/widget_service.py
+++ b/met-api/src/met_api/services/widget_service.py
@@ -52,15 +52,15 @@ class WidgetService:
         return sort_index
 
     @staticmethod
-    def sort_widget(engagement_id, widget_items: list, user_id=None):
+    def sort_widget(engagement_id, widgets: list, user_id=None):
         """Sort widgets."""
-        WidgetService._validate_widget_ids(engagement_id, widget_items)
+        WidgetService._validate_widget_ids(engagement_id, widgets)
 
         widget_sort_mappings = [{
             'id': widget_item.get('id'),
             'sort_index': widget_item.get('sort_index'),
             'updated_by': user_id
-        } for widget_item in widget_items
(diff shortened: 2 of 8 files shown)


met-api/src/met_api/services/widget_service.py
```diff
@@ -32,9 +34,48 @@ def create_widget(widget_data, engagement_id, user_id):
...
+        # find the highest sort order of the engagement
+        sort_index = 0
+        widgets = WidgetModel.get_widgets_by_engagement_id(engagement_id)
+        if widgets:
+            # Find the largest in the existing widgest
+            sort_index = max(widget.sort_index for widget in widgets)
+        return sort_index
+
+    @staticmethod
+    def sort_widget(engagement_id, widget_items: list, user_id=None):
```
jadmsaadaot, line 55: I don't think this is still widget_items it's widgets. right?

met-api/tests/unit/api/test_widget.py
```diff
@@ -34,6 +36,81 @@ def test_create_widget(client, jwt, session, widget_info):  # pylint:disable=unu
                      headers=headers, content_type='application/json')
     assert rv.status_code == 200
 
+    rv = client.get('/api/widgets/engagement/' + str(engagement.id),
+                    headers=headers, content_type='application/json')
+    assert rv.status_code == 200
+    assert rv.json.get('result')[0].get('sort_index') == 1
+
+
+def test_create_widget_sort(client, jwt, session):  # pylint:disable=unused-argument
```
jadmsaadaot, line 45: Too many scenarios in one test? Can they be separated? Failures here is harder to track to what scenario failed. One  scenario for each test preferred

jadmsaadaot (changes requested):
Variable naming issue in widget service. And unit test design issue

met-api/src/met_api/services/widget_service.py
saravanpa-aot: its widget , we are sorting widget

met-api/tests/unit/api/test_widget.py
saravanpa-aot: I can separate the invalid test to another one.. the other test just asserts the sorting order

met-api/migrations/versions/e37d79be3a05_added_sort_for_widgets.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'e37d79be3a05'
+down_revision = '9f86fdcfb248'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### Add Sort index to widget.###
+    op.add_column('widget', sa.Column('sort_index', sa.Integer(), nullable=False, server_default='1'))
```
jadmsaadaot, line 21: learning from this that I don't have to do nullable True, execute then do nullable false. Thanks

met-api/src/met_api/services/widget_service.py
jadmsaadaot: so we need to change the paramater name from widget_items to widgets

tiagograf-AOT (approved):

met-api/src/met_api/services/widget_service.py
saravanpa-aot: yeah..i misunderstood..now its fixed

djnunez-aot (approved):

VineetBala-AOT (approved):

met-api/src/met_api/services/widget_service.py
```diff
@@ -32,9 +34,48 @@ def create_widget(widget_data, engagement_id, user_id):
...
+            'id': widget_item.get('id'),
+            'sort_index': widget_item.get('sort_index'),
+            'updated_by': user_id
+        } for widget_item in widgets
+        ]
+
+        WidgetModel.update_widgets(widget_sort_mappings)
+
+    @staticmethod
+    def _validate_widget_ids(engagement_id, widget_items):
```
jadmsaadaot, line 69: still calling them widget_items here

met-api/tests/unit/api/test_widget.py
jadmsaadaot: We are posting widget_info_1 and widge_info_2 here. We don't need to do that right?

If we are testing the sorting, then the sorting endpoint should be the only one called. The test has multiple API calls, it is a bit difficult to understand what it is testing. Ideally, One test is testing only one API call. Everything else should be given.

met-api/src/met_api/services/widget_service.py
saravanpa-aot: Fixed now

jadmsaadaot (approved):

met-api/tests/unit/api/test_widget.py
saravanpa-aot: Will work on this later..I will create a ticket for tests in model layer and separate the tests

Issue #885, opened 2022-11-18
As a user of MET, I want to add a EAO Phase widget to my engagement


marieanik:
User is on the create engagement page, clicks the "Add widget" button, selects the Engagement Phases widget from the widget selection.

https://xd.adobe.com/view/d0671e99-d506-4ed5-b51e-2bf6e3fcb523-8534/screen/915ff557-feba-41ce-aa36-bed4d306e152

User sees the EA Process widget panel
Title: The EA Process (editable)
Drop-down: Engagement phase

Users selects an engagement phase (Early Engagement, Process Planning, Application Development & Review, Recommendation)

Or checks the "This engagement is a stand-alone engagement checkbox"

https://xd.adobe.com/view/d0671e99-d506-4ed5-b51e-2bf6e3fcb523-8534/screen/42779dff-f05b-48ee-a5b9-8cbd9576e

Clicks Save & Close

Widget is added to engagement

PR #931, opened 2022-11-29
Add fixed and sortable widgets

tiagograf-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/885

*Description of changes:*
Makes engagement phases widget fixed at the top.

### DIFF
--- a/met-web/src/components/engagement/form/EngagementWidgets/WidgetsBlock.tsx
+++ b/met-web/src/components/engagement/form/EngagementWidgets/WidgetsBlock.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState, useCallback } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { Divider, Grid, Skeleton } from '@mui/material';
 import { MetHeader2, MetPaper, SecondaryButton } from 'components/common';
 import { WidgetCardSwitch } from './WidgetCardSwitch';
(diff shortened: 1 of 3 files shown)


met-web/src/components/engagement/form/EngagementWidgets/WidgetsBlock.tsx
```diff
@@ -1,26 +1,31 @@
...
     const { savedEngagement } = useContext(ActionContext);
     const dispatch = useAppDispatch();
 
-    const [tempWidgets, setTempWidgets] = useState<Widget[]>(widgets);
+    const [sortableWidgets, setSortableWidgets] = useState<Widget[]>([]);
+    const [fixedWidgets, setFixedWidgets] = useState<Widget[]>([]);
 
     useEffect(() => {
-        setTempWidgets(widgets);
+        const widgetsFixedList = widgets.filter((w) => w.widget_type_id == WidgetType.Phases);
```
jadmsaadaot, line 25: Maybe we can just define an array of sortable types [WidgetType.Phases, ..] and just check if type id is in it. This way we can just extend the array if mor fixed types existed

tiagograf-AOT: Done

met-web/src/components/engagement/form/EngagementWidgets/WidgetsBlock.tsx
```diff
@@ -33,18 +39,19 @@ const WidgetsBlock = () => {
         handleWidgetDrawerOpen(true);
     };
 
-    const moveWidget = useCallback((dragIndex: number, hoverIndex: number) => {
-        setTempWidgets((prevWidgets: Widget[]) => {
+    const moveWidget = (dragIndex: number, hoverIndex: number) => {
```
tiagograf-AOT, line 42: Not sure what was the reason to use a callback function here, but I had to remove it to have access to fixedWidgets

djnunez-aot (approved):

jadmsaadaot (approved):
Approved, just a bit concerned that without useCallback there might be more calls happening. A refactor for later might be in the backend where simplify the validation. Maybe make sort_index nullable.

Issue #787, opened 2022-10-25
Optimize fetching API Calls for widgets and widget items using Redux RTK


jadmsaadaot:

tiagograf-AOT:
verified and closed

PR #987, opened 2022-12-09
Add RTK Query for documents, widgets, contacts fetches

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/787

*Description of changes:*

- Add redux RTK Query for documents fetch
- Add redux RTK Query for widgets fetch
- Add redux RTK Query for contacts fetch

### DIFF
--- a/met-web/src/components/engagement/form/EngagementWidgets/WhoIsListening/WhoIsListeningContext.tsx
+++ b/met-web/src/components/engagement/form/EngagementWidgets/WhoIsListening/WhoIsListeningContext.tsx
@@ -4,6 +4,7 @@ import { openNotification } from 'services/notificationService/notificationSlice
 import { ActionContext } from '../../ActionContext';
 import { getContacts } from 'services/contactService';
 import { Contact } from 'models/contact';
+import { useLazyGetContactsQuery } from 'apiManager/apiSlices/contacts';
 
 export interface WhoIsListeningContextProps {
     contactToEdit: Contact | null;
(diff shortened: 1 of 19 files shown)


met-web/src/components/engagement/form/EngagementWidgets/WhoIsListening/WhoIsListeningContext.tsx
```diff
@@ -57,7 +59,7 @@ export const WhoIsListeningProvider = ({ children }: { children: JSX.Element | J
                 return Promise.resolve([]);
             }
             setLoadingContacts(true);
-            const loadedContacts = await getContacts();
+            const loadedContacts = await getContactsTrigger(undefined, false).unwrap();
```
tiagograf-AOT, line 62: I am a bit weirded out about passing undefined to the method parameter, hopefully it’s for s good reason?

tiagograf-AOT (approved):
Looks pretty nice, good job!

met-web/src/components/engagement/form/EngagementWidgets/WhoIsListening/WhoIsListeningContext.tsx
jadmsaadaot: yeah you're right. When you define any parameters that the method should take like  (widgetId, engagementId, something) the trigger function will have its first parameter defined as type void. I necessarily needed the second parameter to be false to ignore the cache in this case but couldn't find a value that it accepts other than undefined for the first parameter. Happy to do some more research around this
image

Issue #714, opened 2022-10-07
As a MET system user, I want to see all the current system users on the user management page so that access their user profiles.


marieanik:
https://xd.adobe.com/view/7d5f6547-2523-484d-8101-a7587f9076c8-280c/screen/5c7062ac-642a-4202-b87b-c00bf47b613b/

tiagograf-AOT:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/714?planning-poker) with Zenhub @djnunez-aot @jadmsaadaot @saravanpa-aot @VineetBala-AOT

PR #996, opened 2022-12-13
Added pagination to get endpoints

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/met-public/issues/714

*Description of changes:*

- populated more fields in user table
- Added pagination to get endpoints
- Added tests

### DIFF
--- a/met-api/src/met_api/services/user_service.py
+++ b/met-api/src/met_api/services/user_service.py
@@ -1,4 +1,5 @@
 """Service for user management."""
+from met_api.models.pagination_options import PaginationOptions
 from met_api.models.user import User as UserModel
 from met_api.services.keycloak import KeycloakService
 from met_api.schemas.user import UserSchema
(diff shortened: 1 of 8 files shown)


met-api/src/met_api/services/user_service.py
```diff
@@ -24,13 +25,16 @@ def create_or_update_user(self, user: UserSchema):
         external_id = user.get('external_id')
         db_user = UserModel.get_user_by_external_id(external_id)
         if db_user is None:
+            is_staff_user = user.get('email_id', '').endswith('gov.bc.ca')
```
tiagograf-AOT, line 28: I don't think this will do, lets say I'm from ministry of health and used my gov email to answer a survey, this will think i'm a staff user, the way we are handling this for now is using the external_id which is the idir Id in this case, if null then its public. We might have to have a flag for this in the future

VineetBala-AOT (approved):

tiagograf-AOT (approved):

PR #1002, opened 2022-12-14
Fixes users access type

tiagograf-AOT:
*Issue #:*

*Description of changes:*
Fixes users access type

### DIFF
--- a/met-api/migrations/versions/6764af39864e_.py
+++ b/met-api/migrations/versions/6764af39864e_.py
@@ -1,0 +1,34 @@
+"""
+
+Revision ID: 6764af39864e
+Revises: ffac8f5b4288
+Create Date: 2022-12-13 19:47:27.460920
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+from met_api.utils.enums import UserType
+
+# revision identifiers, used by Alembic.
+revision = '6764af39864e'
+down_revision = 'ffac8f5b4288'
+branch_labels = None
+depends_on = None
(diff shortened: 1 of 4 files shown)


VineetBala-AOT (approved):

met-api/migrations/versions/6764af39864e_.py
```diff
@@ -0,0 +1,34 @@
...
+    # reset users type
+    update_reset_access_type_query = f"update met_users set access_type = null where 0=0"
+    op.execute(update_reset_access_type_query)
+
+    # update IDIR users to type STAFF
+    username = 'idir'
+    update_staff_type_query = f"update met_users set access_type = '{UserType.STAFF.value}' where username like '%{username}'"
+    op.execute(update_staff_type_query)
+    # ### end Alembic commands ###
+
```
saravanpa-aot, line 30: I had the same thing in previous PR ; did i miss anything?

saravanpa-aot (approved):

met-api/migrations/versions/6764af39864e_.py
tiagograf-AOT: i'm using the username instead, I thought this caused another problem, but i found out later that the.root cause was the update, i'm keeping the migration since it seems to work based on usernname

saravanpa-aot: yeah..cool..go ahead ..we will change it in the next PR properly...

Issue #402, opened 2022-08-09
The only mandatory fields to create an engagement should be Engagement Name and Date


marieanik:
As a Engagement Specialist I want to create an engagement only using its name and dates at first so that I can enter a description and content in the future.

Acceptance Criteria

Create/Update engagement only required fields are Name and Date
Preview Engagement banner shows missing description and/or content.

tiagograf-AOT:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/402?planning-poker) with Zenhub @carlygreenehill @djnunez-aot @jadmsaadaot @marieanik @saravanpa-aot @VineetBala-AOT

marieanik:
@djnunez-aot Just tested in Dev andf still getting the "Content missing" error message.![Screen Shot 2022-12-16 at 7.28.22 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/e7c4c009-5421-4bb8-8734-271bf6386462)

djnunez-aot:
@marieanik The pr is not merged yet so it is still only on my local I will ping you once it is ready for testing

PR #1025, opened 2022-12-16
Remove rich description & content as mandatory fields for Engagement Form

djnunez-aot:
*Issue #402 :*
https://github.com/bcgov/met-public/issues/402

- Remove description and content as mandatory fields for engagement creation
-Add indicator in preview page that content and description is missing
-Prevent scheduling if engagement has no description or content
-Update back end by removing description and content as required fields for db

### DIFF
--- a/met-web/src/components/common/Modals/Schedule/index.tsx
+++ b/met-web/src/components/common/Modals/Schedule/index.tsx
@@ -23,11 +23,16 @@ const ScheduleModal = ({ reschedule, open, updateModal }: ScheduleModalProps) =>
     const [scheduledDate, setScheduledDate] = useState<Dayjs | null>(dayjs(Date.now()));
     const { savedEngagement, scheduleEngagement } = useContext(ActionContext);
     const dispatch = useAppDispatch();
-    const isValid =
-        savedEngagement.surveys.length === 0 &&
-        savedEngagement.content &&
-        savedEngagement.description &&
-        savedEngagement.banner_url;
+
+    const isEngagementReady = () => {
+        return (
+            savedEngagement.surveys.length === 0 &&
+            savedEngagement.content &&
+            savedEngagement.description &&
+            savedEngagement.banner_url
+        );
+    };
+
     const handleChange = (newDate: Dayjs | null) => {
         if (newDate != null) {
             setScheduledDate(newDate);
@@ -49,7 +54,7 @@ const ScheduleModal = ({ reschedule, open, updateModal }: ScheduleModalProps) =>
     };
 
     const handleSchedule = async () => {
-        if (!isValid) {
+        if (!isEngagementReady()) {
             dispatch(
                 openNotification({
                     severity: 'error',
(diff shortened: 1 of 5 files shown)


met-web/src/components/common/Modals/Schedule/index.tsx
```diff
@@ -23,7 +23,11 @@ const ScheduleModal = ({ reschedule, open, updateModal }: ScheduleModalProps) =>
     const [scheduledDate, setScheduledDate] = useState<Dayjs | null>(dayjs(Date.now()));
     const { savedEngagement, scheduleEngagement } = useContext(ActionContext);
     const dispatch = useAppDispatch();
-
+    const isValid =
```
tiagograf-AOT, line 26: I think it should be called something like IsEngagementReady
but anyway can you put this in a separated validation method?

met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -217,7 +211,7 @@ const EngagementForm = () => {
                     columnSpacing={2}
                 >
                     <Grid item xs={12}>
-                        <MetLabel>Engagement Date</MetLabel>
+                        <MetLabel>Engagement Date {isNewEngagement ? '*' : ''}</MetLabel>
```
tiagograf-AOT, line 214: This is always required so i think you can leave the * there at all times

met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -188,7 +182,7 @@ const EngagementForm = () => {
                     />
                 </Grid>
                 <Grid item xs={12} lg={8} md={6}>
-                    <MetLabel sx={{ marginBottom: '2px' }}>Engagement Name</MetLabel>
+                    <MetLabel sx={{ marginBottom: '2px' }}>Engagement Name {isNewEngagement ? '*' : ''}</MetLabel>
```
tiagograf-AOT, line 185: same here, * visible at all times

met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -79,7 +77,6 @@ const EngagementForm = () => {
 
         setEngagementFormError({
```
tiagograf-AOT, line 78: I am assuming we can drop this call all together since the description doesn't affect the formerror anymore

met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -91,7 +88,6 @@ const EngagementForm = () => {
...
```
tiagograf-AOT, line 89: same here

met-web/src/components/common/Modals/Schedule/index.tsx
tiagograf-AOT: one other. thing that I thought is.  that. we validate that to schedule the engagement, but the user. is still able to edit it after its been schedule, So in teory i can schedule the engagement and then remove the description, we might nneed to confirm with marie anik if this is okay

jadmsaadaot (approved):

saravanpa-aot (changes requested):
I talked to @marieanik and she thinks we don't need * 
Plz bring this up in the stand up and we can make the changes then

saravanpa-aot (approved):

Issue #1001, opened 2022-12-14
Edit Document moves it to the bottom


tiagograf-AOT:
Steps to Reproduce:

Edit a document or folder in the documents widget
After saving the document structure is rearranged and the edited document/folder is sent to the bottom.

Expected:

After saving. the document/folder should stay where it was

VineetBala-AOT:
Tested the edited documents or folder structure is not getting rearranged now.

PR #1033, opened 2022-12-19
Fix document reordering

djnunez-aot:
*Issue #1001 :*
https://github.com/bcgov/met-public/issues/1001

Expectation:
-When document folders or files are edited do not reorder the data files randomly.

Current version: 
-When documents are reordered they randomly shuffle order.

This pr:
-Sort folders by ascending id order to  make document order consistent in the widget tab 

### DIFF
--- a/met-api/src/met_api/services/widget_documents_service.py
+++ b/met-api/src/met_api/services/widget_documents_service.py
@@ -29,6 +29,7 @@ class WidgetDocumentService:
     @staticmethod
     def _attach_file_nodes(docs, root):
         files = list(filter(lambda doc: doc.type == DocumentType.FILE.value, docs))
+        files.sort(key=lambda doc: doc.id)
         for file in files:
             props = WidgetDocumentService._fetch_props(file)
             parent_node = root
@@ -38,7 +39,9 @@ class WidgetDocumentService:
 
     @staticmethod
     def _attach_folder_nodes(docs, root):
+
         folders = list(filter(lambda doc: doc.type == DocumentType.FOLDER.value, docs))
+        folders.sort(key=lambda doc: doc.id)
         for folder in folders:
             props = WidgetDocumentService._fetch_props(folder)
             AnyNode(**props, parent=root)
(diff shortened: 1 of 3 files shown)


saravanpa-aot (changes requested):
Can you improve the PR description to state the problem that you are handling ?
Or add a test to reflect these changes?
Without both of these , its hard to know what exactly this change does

met-api/src/met_api/services/widget_documents_service.py
```diff
@@ -29,6 +29,7 @@ def get_documents_by_widget_id(widget_id):
     @staticmethod
     def _attach_file_nodes(docs, root):
         files = list(filter(lambda doc: doc.type == DocumentType.FILE.value, docs))
+        files.sort(key=lambda doc: doc.id)
```
jadmsaadaot, line 32: is there a requirement to sort by id?

tiagograf-AOT (approved):

met-api/src/met_api/services/widget_documents_service.py
djnunez-aot: yes without sorting by id for the files, when there is multiple files without a folder they randomly change order after being edited.

jadmsaadaot (approved):

met-api/tests/unit/models/test_widget_documents.py
```diff
@@ -55,6 +55,7 @@ def test_get_all_by_widget_id(session):
     session.commit()
 
     expected_docs = WidgetDocumentsModel.get_all_by_widget_id(widget.id)
+    all(expected_docs[i].id <= expected_docs[i + 1].id for i in range(len(expected_docs) - 1))
```
saravanpa-aot, line 58: what does this do?

djnunez-aot: it checks if the expected docs are order in ascending order

saravanpa-aot (approved):

PR #1036, opened 2022-12-19
Allo multi page form submission and handle extract comments

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/978

*Description of changes:*

- Enable submission for multi-page survey
- handle comments extraction

### DIFF
--- a/met-api/src/met_api/services/comment_service.py
+++ b/met-api/src/met_api/services/comment_service.py
@@ -95,7 +95,6 @@ class CommentService:
         """Extract comments from survey submission."""
         survey_form = survey.get('form_json', {})
         components = cls.extract_components(survey_form)
-        print(components)
         if len(components) == 0:
             return []
         # get the 'id' for each component that has 'inputType' text and filter out the rest.
@@ -105,5 +104,4 @@ class CommentService:
         submission = survey_submission.get('submission_json', {})
         comments = [cls.__form_comment(key, submission.get(key, ''), survey_submission, survey)
                     for key in text_component_keys if submission.get(key, '') != '']
-        print(comments)
         return comments
--- a/met-api/src/met_api/services/comment_service.py
+++ b/met-api/src/met_api/services/comment_service.py
@@ -13,7 +13,7 @@ class CommentService:
     """Comment management service."""
 
     otherdateformat = '%Y-%m-%d'
-    
+
     wizard_display = 'wizard'
     form_display = 'form'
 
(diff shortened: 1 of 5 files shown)


met-api/src/met_api/services/comment_service.py
```diff
@@ -73,11 +78,23 @@ def __form_comment(component_id, comment_text, survey_submission: SubmissionSche
             'submission_id': survey_submission.get('id', None)
         }
 
+    @classmethod
+    def extract_components(cls, survey_form: dict):
+        """Extract components from survey form."""
+        components = list(survey_form.get('components', []))
+
+        if survey_form.get('display') == cls.form_display:
+            return components
```
tiagograf-AOT, line 87: So if we ever enable a panel or "container" components this won't work right?

tiagograf-AOT (approved):

met-api/src/met_api/services/comment_service.py
tiagograf-AOT: I wonder if our code should always check for iterable components

met-api/src/met_api/services/comment_service.py
```diff
@@ -73,11 +78,23 @@ def __form_comment(component_id, comment_text, survey_submission: SubmissionSche
             'submission_id': survey_submission.get('id', None)
         }
 
+    @classmethod
+    def extract_components(cls, survey_form: dict):
+        """Extract components from survey form."""
+        components = list(survey_form.get('components', []))
+
+        if survey_form.get('display') == cls.form_display:
```
saravanpa-aot, line 86: a few comments about logic will be helpful ; or some tests 
comments will be easy i guess..

met-api/src/met_api/services/comment_service.py
```diff
@@ -73,11 +78,23 @@ def __form_comment(component_id, comment_text, survey_submission: SubmissionSche
...
+            return components
+
+        if survey_form.get('display') == cls.wizard_display:
```
saravanpa-aot, line 89: one another way to avoid comment will be 
```
is_comments_from_wizard_survey = survey_form.get('display') == cls.wizard_display:
if is_comments_from_wizard_survey:
```

saravanpa-aot (approved):

met-api/src/met_api/services/comment_service.py
jadmsaadaot: No, this code does not cover that scenario, but it's not in scope for this task. This is a simple solution just to allow for the multi-page submission. To check for nested iterable components to allow usage of layout components in survey is more complex and better handled separately.

met-api/src/met_api/services/comment_service.py
jadmsaadaot: done

met-api/src/met_api/services/comment_service.py
jadmsaadaot: done

PR #1041, opened 2022-12-21
Edit submission front end

tiagograf-AOT:
*Issue #:*

*Description of changes:*
New public route/form to edit submission


djnunez-aot (approved):

met-web/src/components/survey/edit/ActionContext.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    severity: 'error',
+                    text: 'Verification token is invalid.',
+                }),
+            );
+            setTokenValid(false);
+        }
+    };
+
+    const loadSubmission = async () => {
+        if (token && isTokenValid)
```
jadmsaadaot, line 83: I would consider returning early here rather than wrapping everything with the if. if(!token || !isTokenValid) return;. And rest of logic is independent of it.

met-web/src/components/survey/edit/ActionContext.tsx
```diff
@@ -0,0 +1,169 @@
...
+                openNotification({
+                    severity: 'error',
+                    text: 'Error occurred while loading engagement',
+                }),
+            );
+        }
+    };
+
+    const handleSubmit = async () => {
+        if (token)
```
jadmsaadaot, line 122: same suggestion to return early. if(!token) return;

met-web/src/components/survey/edit/EditForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+                            multiline={true}
+                            rows={3}
+                            onChange={(e) => {
+                                comment.text = e.target.value;
+                            }}
+                        />
+                    </Grid>
+                );
+            })}
+            <Grid item container xs={12} direction="row" justifyContent="flex-end" spacing={1} sx={{ mt: '1em' }}>
```
jadmsaadaot, line 35: Have we considered mobile view here. If we needed to switch Secondary and Primary button positions for mobile, I've discovered this neat order property. Hope it works
image

met-web/src/components/survey/edit/EditForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+                return (
+                    <Grid item xs={12} key={index}>
+                        <MetLabel>{comment.label}</MetLabel>
+                        <TextField
+                            defaultValue={comment.text}
+                            sx={{ width: '100%' }}
+                            multiline={true}
+                            rows={3}
+                            onChange={(e) => {
+                                comment.text = e.target.value;
```
jadmsaadaot, line 29: I feel weird about handling state change like this. I think it could be against react pattern because state change should always occur using setState hook or this.setState for class components. Maybe we could pass the setSubmission and use the index to update the particular comment?

met-web/src/components/survey/edit/EditForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+                            }}
+                        />
+                    </Grid>
+                );
+            })}
+            <Grid item container xs={12} direction="row" justifyContent="flex-end" spacing={1} sx={{ mt: '1em' }}>
+                <Stack
+                    direction={{ md: 'column-reverse', lg: 'row' }}
```
jadmsaadaot, line 37: very neat

met-web/src/components/survey/edit/InvalidTokenModal.tsx
```diff
@@ -0,0 +1,45 @@
...
+                </Grid>
+                <Grid item xs={12}>
+                    <MetBody sx={{ p: '1em', borderLeft: 8, borderColor: '#003366', backgroundColor: '#F2F2F2' }}>
+                        This link has already been previously used. Or
+                        <br />
+                        The engagement period is over.
+                    </MetBody>
+                </Grid>
+                <Grid item container xs={12} justifyContent="flex-end" spacing={1} sx={{ mt: '1em' }}>
+                    <PrimaryButton onClick={handleClose}>Back to Engagement</PrimaryButton>
```
jadmsaadaot, line 40: is the behavior the same when user clicks away, like backdrop click

jadmsaadaot (changes requested):
state update without using setState

met-web/src/components/survey/edit/EditForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const EditForm = ({ handleClose }: SurveyFormProps) => {
+    const { handleSubmit, isSubmitting, submission, setSubmission } = useContext(ActionContext);
+
+    const handleChange = (value: string, commentIndex: number) => {
+        if (!submission) {
+            return;
+        }
+
+        submission.comments[commentIndex].text = value;
```
jadmsaadaot, line 15: React's convention is that we should always get create a copy of the state if we wanted to mutate it. Check this out https://stackoverflow.com/a/40309023

jadmsaadaot (approved):

PR #1053, opened 2022-12-28
Update Email Verification Expiration Logic

tiagograf-AOT:
*Issue #:*

*Description of changes:*
Changes to validating the email verification expiration logic

### DIFF
--- a/met-api/src/met_api/services/email_verification_service.py
+++ b/met-api/src/met_api/services/email_verification_service.py
@@ -34,7 +34,7 @@ class EmailVerificationService:
         """Get an active email verification matching the provided token."""
         db_email_verification = EmailVerification.get(verification_token)
         email_verification = EmailVerificationSchema().dump(db_email_verification)
-        cls.validate_object(email_verification)
+        cls.validate_email_verification(email_verification)
         return email_verification
 
     @classmethod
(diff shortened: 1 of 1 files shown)


met-api/src/met_api/services/email_verification_service.py
```diff
@@ -131,12 +131,14 @@ def validate_object(email_verification: EmailVerificationSchema):
         if not is_active:
             raise ValueError('Email verification already verified')
 
-        verification_created_datetime = datetime.strptime(
-            email_verification.get('created_date'), EmailVerificationService.datetime_format)
-        verification_expiry_datetime = verification_created_datetime + \
-            timedelta(hours=EmailVerificationService.verification_expiry_hours)
-        if datetime.now() >= verification_expiry_datetime:
-            raise ValueError('Email verification is expired')
+        if email_verification.get('submission_id') is None:
```
tiagograf-AOT, line 134: We should likely come up with a more permanent solution for expiring email verification tokens (likely a new expiration date column) but it has a few tricky things about it so I'm going with the simple fix for now

saravanpa-aot: afaik , we shouldnt expire the email token till the submission is open rite?then we can check the expiry agianst the engagement schedule date?

tiagograf-AOT: thats right, it does check the engagement dates right below, this change will bypass the check for the 24 h expiration date if the token was generated for editing a submission

tiagograf-AOT: i didn't want to have a specific expiration date set for the engagement dates since we still allow the engagement dates to be changed while its active

jadmsaadaot (approved):

saravanpa-aot (approved):

Issue #1016, opened 2022-12-15
Review comments updates


marieanik:
As a MET user, I want more options to review and reject comments.

Add the option "Needs further review" below "Rejected"

##Default View##

Add a text area for internal note. These notes are editable and will be visible only on the comment review page.

Label: Internal Note

***EDIT FROM CARLY: Adding an internal note, or updating existing note, SHOULD NOT trigger another email to public user re: rejected comment. ie. saving a new internal note, hitting save, should not send another reject email to public user***

***Internal Note is ALWAYS visible, regardless of approve/rejected***

https://xd.adobe.com/view/7d5f6547-2523-484d-8101-a7587f9076c8-280c/screen/d74f79f5-7cde-46eb-bd3e-2b3ad6258d9d/

##Rejected##

If rejected option is selected, add the following:

Text area for Review Note
Label: Review Note _(this note will be inserted in the email sent to the respondent to help them understand what needs to be edited for their comment(s) to be approved.)_
Button: Preview Email (https://app.zenhub.com/workspaces/met-ops-6254279dcfa60b0010ea5a0d/issues/gh/bcgov/met-public/1017)

Add a text area for internal note. These notes are editable and will be visible only on the comment review page.

Label: Internal Note

Then, between the text "Clicking the “Save” button will trigger an automatic email to be sent to the person who made this comment. They will have to option to edit and re-submit their comment.
The edited comment will have to be approved before it is published." and the "Threat" section, add a checkbox:

Label: Don't send this email to the person who commented.

If this checkbox is checked, no email is sent. 

https://xd.adobe.com/view/7d5f6547-2523-484d-8101-a7587f9076c8-280c/screen/90332d54-a3fd-41e6-a174-36a4d976c126/

Make sure all the spacings are like on the wireframe. 

![Screen Shot 2022-12-15 at 5.35.14 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/ea1e0617-284b-493c-998a-05f4a7ed00e1)

***EDIT FROM CARLY: If "Needs Review" is selected, Comment status remains as PENDING***

carlygreenehill:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/1016?planning-poker) with Zenhub @jadmsaadaot @djnunez-aot @VineetBala-AOT @saravanpa-aot

[...]

PR #1090, opened 2023-01-11
back end changes for review comments updates

VineetBala-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/1016

*Description of changes:*
- Added columns and table to handle notify email and Review/Internal Note and logic to store the details
- Changed the logic to handle rejection email only in case of changes on review comment or rejection reason


met-api/migrations/versions/0e043f976e2e_add_notes_to_submission.py
```diff
@@ -0,0 +1,38 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('submission', sa.Column('notify_email', sa.Boolean(), nullable=True))
+    comment_status_table = sa.table('comment_status',
+        sa.Column('id', sa.Integer()),
+        sa.Column('status_name', sa.String(length=50)),
+        sa.Column('description', sa.String(length=50)),
+        sa.Column('created_date', sa.DateTime()),
+        sa.Column('updated_date', sa.DateTime()))
+    op.bulk_insert(comment_status_table, [
+        {'id': 4, 'status_name': 'Needs further review', 'description': 'Comment needs further review', 'created_date': datetime.utcnow(), 'updated_date': datetime.utcnow()},
```
tiagograf-AOT, line 30: I mentioned this in the other PR i don't think we need a new status. we can chat about this

met-api/src/met_api/models/staff_note.py
```diff
@@ -0,0 +1,61 @@
...
+            .all()
+
+    @classmethod
+    def update_staff_note(cls, staff_note: dict, session=None) -> StaffNote:
+        """Update existing staff note."""
+        note_id = staff_note.get('id', None)
+        query = StaffNote.query.filter_by(id=note_id)
+
+        update_note = dict(
+            note=staff_note.get('note', None),
```
tiagograf-AOT, line 52: make sure to update the updated_by updated_date columns here

tiagograf-AOT: also, i can't find the add/create method anywhere?

VineetBala-AOT: Verified the columns updated_by and updated_date. Its getting updated on record update. Add and update is happening in add_or_update_staff_note

met-api/src/met_api/services/submission_service.py
```diff
@@ -140,6 +152,75 @@ def validate_review(cls, values: dict, user):
...
             raise ValueError('A rejection reason is required.')
 
+    @classmethod
+    def add_or_update_staff_note(cls, survey_id, submission_id, staff_notes):
+        """Process staff note for a comment."""
+        for staff_note in staff_notes:
+            note = StaffNote.get_staff_note_by_type(submission_id, staff_note.get('note_type'))
+            if note:
+                note.note = staff_note['note']
+                StaffNote.flush(note)
```
tiagograf-AOT, line 162: flush a save method from the basemodel right? I think its the fist time i've seen it in our repo so i'm confused

met-api/src/met_api/services/submission_service.py
```diff
@@ -140,6 +152,75 @@ def validate_review(cls, values: dict, user):
...
+        #   if review note has not changed
+
+        if staff_comment_details.get('has_threat') is True:
+            return False
+        if staff_comment_details.get('notify_email') is False:
+            return False
+        if staff_comment_details.get('status_id') == Status.Rejected.value:
+            return True
+        review_note_changed = SubmissionService.is_review_note_changed(submission_id, staff_comment_details)
+        if review_note_changed:
```
tiagograf-AOT, line 193: I'm not sure about this logic, 
if the comment status is rejected it will always send the email (except for the two conditions above)
if its not rejected (lets say needs further review) and there's a change in the review notes (which I think it is only visible when status = rejected) it will send an email
shouldn't only send when its rejected?

saravanpa-aot (approved):

met-api/src/met_api/services/submission_service.py
VineetBala-AOT: changed it to check rejection reason changed within rejection before sending email.

met-api/src/met_api/services/submission_service.py
```diff
@@ -140,6 +152,75 @@ def validate_review(cls, values: dict, user):
...
+        # Send the mail
+        #   if the status of the comment is rejected
+        #   if review note has not changed
+
+        if staff_comment_details.get('has_threat') is True:
+            return False
+        if staff_comment_details.get('notify_email') is False:
+            return False
+        if staff_comment_details.get('status_id') == Status.Rejected.value:
+            return True
```
tiagograf-AOT: i think you need to remove the return on this line, this will return True if status == rejected and it won't check if the note has changed e.g.

if status == rejected AND note_changed
 return true

tiagograf-AOT (approved):

met-api/src/met_api/models/staff_note.py
```diff
@@ -0,0 +1,61 @@
...
+    note_type = db.Column(db.String(50), nullable=True)
+    survey_id = db.Column(db.Integer, ForeignKey('survey.id', ondelete='CASCADE'), nullable=False)
+    submission_id = db.Column(db.Integer, ForeignKey('submission.id', ondelete='SET NULL'), nullable=False)
+
+    @classmethod
+    def get_staff_note(cls, note_id):
+        """Get staff note."""
+        return db.session.query(StaffNote)\
+            .filter(StaffNote.id == note_id)\
+            .all()
```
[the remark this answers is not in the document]
VineetBala-AOT: Thanks removed it.

met-api/src/met_api/services/submission_service.py
VineetBala-AOT: Thanks have changed it to check the review not or reasons within this.

PR #1091, opened 2023-01-11
front end changes for review comments updates

VineetBala-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/1016

*Description of changes:*
- Changes made to handle internal note for any comment review reason3
- Changes made to handle rejection review comments
- Added additional review reason
- Added the option to not send rejection email

### DIFF
--- a/met-web/src/components/comments/admin/review/CommentReview.tsx
+++ b/met-web/src/components/comments/admin/review/CommentReview.tsx
@@ -152,10 +152,13 @@ const CommentReview = () => {
         return <CommentReviewSkeleton />;
     }
 
-    const handleNoteChange = (note: string, index: number) => {
+    const handleNoteChange = (note: string, note_type: string, note_id: number) => {
         const newStaffNoteArray = [...updatedStaffNote];
-        newStaffNoteArray[index].note = note;
-        newStaffNoteArray[index].survey_id = submission.survey_id;
+        newStaffNoteArray.map((staffNote) => {
+            if (staffNote.id === note_id && staffNote.note_type === note_type) {
+                staffNote.note = note;
+            }
+        });
         setUpdatedStaffNote(newStaffNoteArray);
     };
 
@@ -335,7 +338,7 @@ const CommentReview = () => {
                                         respondent to help them understand what needs to be edited for their comment(s)
                                         to be approved.)
                                     </MetParagraph>
-                                    {reviewNotes.map((staffNote, index) => {
+                                    {reviewNotes.map((staffNote) => {
                                         return (
                                             <TextField
                                                 value={staffNote.note}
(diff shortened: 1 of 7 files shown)


met-web/src/components/comments/admin/review/CommentReview.tsx
```diff
@@ -72,10 +79,20 @@ const CommentReview = () => {
         }
     };
 
+    const extract_staff_note = async () => {
+        staffNote.length === 0
```
tiagograf-AOT, line 83: I'd use this the other way around setUpdatedStaffNote(staffNote.len... ? staffNote : default..)

met-web/src/constants/commentStatus.ts
```diff
@@ -2,4 +2,5 @@ export enum CommentStatus {
     Pending = 1,
     Approved = 2,
     Rejected = 3,
+    NeedsFurtherReview = 4,
```
tiagograf-AOT, line 5: If I remember correctly I think NeedsFurtherReview = Pending, so i think we don't need a new status

met-web/src/components/comments/admin/review/CommentReview.tsx
```diff
@@ -295,12 +327,71 @@ const CommentReview = () => {
                                         onChange={(event) => setOtherReason(event.target.value)}
                                     />
                                     <br />
+                                    <MetParagraph>
+                                        <b>Review Note</b> (this note will be inserted in the email sent to the
+                                        respondent to help them understand what needs to be edited for their comment(s)
+                                        to be approved.)
+                                    </MetParagraph>
+                                    {updatedStaffNote.map((staffNote, index) => {
```
jadmsaadaot, line 335: Let's create two constant before the return part, const reviewNotes = updatedStaffNote.filter( note => note.type === StaffNoteType[StaffNoteType.Review]. like this then use map to display them here

jadmsaadaot: also what does StaffNoteType[StaffNoteType.Review] return?

jadmsaadaot: consider something like this:
image

jadmsaadaot (approved):

met-web/src/components/comments/admin/review/CommentReview.tsx
VineetBala-AOT: addressed it, thanks.

met-web/src/components/comments/admin/review/CommentReview.tsx
VineetBala-AOT: addressed it, thanks.

tiagograf-AOT (approved):

Issue #680, opened 2022-10-04
Subscribe to an engagement widget


marieanik:
As a MET system user, I want to add a subscribe widget so the public can register to be on the EAO mailing list.

##scenario 1## 
MET system user clicks button to add widget, sees an option to add subscribe widget. 
Name: "Sign Up for Updates"
Description: "Offer members of the public to sign up for updates"
![Screen Shot 2022-12-19 at 10.58.00 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/396b5267-7080-4cd2-a0a0-9bc01cedbc2a)
Selects that option, and widget is automatically added to the widget section. (with a confirmation message)

##scenario 2##
Preview/public view 
Header: Sign Up for Updates
Content: If you are interested in getting updates on public engagements at the EAO, you can sign up below:
Button: Sign Up for Updates from the EAO
![Screen Shot 2022-12-19 at 10.56.37 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/2f0fdf24-9d48-4694-8445-2363b5698e63)
Opens Modal:
Header: Sign Up for Updates
Content: Sign up to receive news and updates on public engagements at the EAO.
FOIPPA and input field for email (same as validate email modal)
Buttons: Cancel & Submit
![Screen Shot 2022-12-19 at 11.01.32 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/ac653723-10a2-4722-8d88-6e9b37f2c83b)

#scenario 3## 
Public user fills in the modal and clicks Submit.

Receive a confirmation message in the modal.

Header: Thank you
Text: We sent a link to confirm your subscription at the following email address:

Bold Text: Please click the link provided to confirm your interest in receiving news and updates from the EAO.

Button: "Close"

![Screen Shot 2022-12-22 at 7.37.16 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/973249b5-e09d-4db7-930a-7eaecc2c2b58)

tiagograf-AOT:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/680?planning-poker) with Zenhub @djnunez-aot @jadmsaadaot @saravanpa-aot @VineetBala-AOT

PR #1117, opened 2023-01-17
updates to subscribe widget & modal sub text

djnunez-aot:
#680 
https://github.com/bcgov/met-public/issues/#680

-Change Sign up to Sign Up

-Make subscribe modal subtext bold:
  To do this we added the subtext prop to all modal subtexts for future use cases

### DIFF
--- a/met-web/src/components/common/Modals/types.tsx
+++ b/met-web/src/components/common/Modals/types.tsx
@@ -1,6 +1,6 @@
 export interface ModalSubtext {
     text: string;
-    bold: boolean;
+    bold?: boolean;
 }
 
 export interface NotificationModalProps {
--- a/met-web/src/components/common/Modals/types.tsx
+++ b/met-web/src/components/common/Modals/types.tsx
@@ -1,6 +1,11 @@
+export interface ModalSubtext {
+    text: string;
+    bold: boolean;
+}
+
 export interface NotificationModalProps {
     header: string;
-    subText: string[];
+    subText: ModalSubtext[];
     handleConfirm?: () => void;
     handleClose?: () => void;
 }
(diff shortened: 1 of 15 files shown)


tiagograf-AOT (approved):

met-web/src/components/common/Modals/types.tsx
```diff
@@ -1,6 +1,11 @@
+export interface ModalSubtext {
+    text: string;
+    bold: boolean;
```
jadmsaadaot, line 3: bold should be an optional field bold?: string. And we don't have to do bold: false when we create a ModalSubtext, we only pass it if it's true

met-web/src/components/engagement/form/AddSurveyBlock.tsx
```diff
@@ -75,8 +75,14 @@ export const AddSurveyBlock = ({
                 data: {
                     header: 'Remove Survey',
                     subText: [
-                        'You will be removing this survey from the engagement. This survey will not be deleted and will be available to add to any engagement.',
-                        'Do you want to remove this survey?',
+                        {
+                            text: 'You will be removing this survey from the engagement. This survey will not be deleted and will be available to add to any engagement.',
+                            bold: false,
```
jadmsaadaot, line 80: we shouldn't pass it if it's false

jadmsaadaot (changes requested):
false parameter should be an optional one

met-web/src/components/engagement/form/AddSurveyBlock.tsx
djnunez-aot: changed

jadmsaadaot (approved):

Issue #984, opened 2022-12-08
Add Contacts doesn't display properly


carlygreenehill:
Instructions on how to load image to Add Contacts has too much padding and sprawls on my screen (browser = Edge, my screen is 15")

![image.png](https://images.zenhubusercontent.com/6261d2829e1fc06fb246914e/5aad60d5-e321-4fd4-a86b-d7875086707d)

PR #1119, opened 2023-01-17
Adjust margin for who is listening widget

djnunez-aot:
*Issue #984 :*
https://github.com/bcgov/met-public/issues/#984

-Small margin adjustments so the add contact drawer has better responsiveness

### DIFF
--- a/met-web/src/components/imageUpload/index.tsx
+++ b/met-web/src/components/imageUpload/index.tsx
@@ -90,7 +90,7 @@ const ImageUpload = ({
                         }}
                     >
                         <input {...getInputProps()} multiple={false} accept={'image/*'} />
-                        <Typography m={4}>{helpText}</Typography>
+                        <Typography m={2}>{helpText}</Typography>
                     </Grid>
                 </section>
             )}


met-web/src/components/imageUpload/index.tsx
```diff
@@ -90,7 +90,7 @@ const ImageUpload = ({
                         }}
                     >
                         <input {...getInputProps()} multiple={false} accept={'image/*'} />
-                        <Typography m={4}>{helpText}</Typography>
+                        <Typography m={2}>{helpText}</Typography>
```
jadmsaadaot, line 93: This also affect the banner image upload. Just make sure that one doesn't look weird now.
 maybe we can pass this margin as a prop

jadmsaadaot (changes requested):
Change affects banner image as well

met-web/src/components/imageUpload/index.tsx
djnunez-aot: I checked both it looks great on both but im open to that as well

djnunez-aot: done

jadmsaadaot (approved):

met-web/src/components/imageUpload/index.tsx
```diff
@@ -4,11 +4,13 @@ import Dropzone from 'react-dropzone';
 import { SecondaryButton } from 'components/common';
 
 interface ImageUploadProps {
+    margin: number;
```
jadmsaadaot, line 7: it's an optional paramter

met-web/src/components/imageUpload/index.tsx
```diff
@@ -4,11 +4,13 @@ import Dropzone from 'react-dropzone';
...
     handleAddFile: (_files: File[]) => void;
     savedImageUrl?: string;
     helpText?: string;
 }
 const ImageUpload = ({
+    margin,
```
jadmsaadaot, line 13: margin has default value. So marge = 2,

jadmsaadaot (changes requested):
make margin optional and has a value by default

Issue #1069, opened 2023-01-04
Change Buttons wording on Engagement creation page


marieanik:
During testing, I discovered there was confusion with the buttons on the Engagement Creation page. 

Users are looking for a "Save" button. Not a single user clicked the Create/Update button. They went straight to the "Save & Preview" (because it says "Save) and were not sure why you couldn't just save without previewing.

All users would expect clicking a "Preview" button would save their work & didn't think it was necessary to include "Save &".

To make it clearer for the user, we will change the text on the buttons for:

Create/Update Engagement > Save
Save & Preview > Preview

Behaviour remains the same.

PR #1131, opened 2023-01-19
Update Engagement Form names

djnunez-aot:
*Issue #1069:*
https://github.com/bcgov/met-public/issues/1069

-Change create/update engagement text to save
-Change preview & save engagement text to preview

### DIFF
--- a/met-web/src/components/engagement/form/EngagementForm.tsx
+++ b/met-web/src/components/engagement/form/EngagementForm.tsx
@@ -306,16 +306,16 @@ const EngagementForm = () => {
                         <Then>
                             <PrimaryButton
                                 sx={{ marginRight: 1 }}
+                                data-testid="create-engagement-button"
                                 onClick={() => handleCreateEngagement()}
                                 loading={isSaving}
-                                data-testid="engagement-form/create-engagement-button"
                             >
                                 Save
                             </PrimaryButton>
                         </Then>
                         <Else>
                             <PrimaryButton
-                                data-testid="engagement-form/update-engagement-button"
+                                data-testid="update-engagement-button"
                                 sx={{ marginRight: 1 }}
                                 onClick={() => handleUpdateEngagement()}
                                 disabled={isSaving}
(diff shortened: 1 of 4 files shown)


met-web/src/components/engagement/form/EngagementForm.tsx
```diff
@@ -306,11 +306,12 @@ const EngagementForm = () => {
                         <Then>
                             <PrimaryButton
                                 sx={{ marginRight: 1 }}
-                                data-testid="engagement-form/create-engagement-button"
                                 onClick={() => handleCreateEngagement()}
                                 loading={isSaving}
                             >
-                                Create Engagement Draft
+                                <React.Fragment data-testid="engagement-form/create-engagement-button">
```
jadmsaadaot, line 312: What makes us have to create a React Fragment so we can pass data test id, can't we keep it in the primary button and get the button by it in the test and check for its existence?

jadmsaadaot (changes requested):
Unnecessary usage of React Fragment

saravanpa-aot (commented):
the title just looks like just label changes ?
But PR has more changes?

met-web/src/components/engagement/form/EngagementForm.tsx
djnunez-aot: Ill remove that I was trying it out since data-testid was still giving me test errors on the Primary button

djnunez-aot: @saravanpa-aot it was label changes originally but I started getting some jest errors so need to update the tests

saravanpa-aot: hm..i think i will get to the bottom of the issue if the test works now in current state and a small text change breaks the tests...

saravanpa-aot: if a text change requires tests to be modified , thats an improvement area for sure..We can discuss this

djnunez-aot: ok sounds good

jadmsaadaot (approved):

saravanpa-aot (approved):

PR #1133, opened 2023-01-20
Common Folder Cleanup

djnunez-aot:
*Issue #1081 :*
https://github.com/bcgov/met-public/issues/1081

-Common Folder Cleanup

-Move style constants


tiagograf-AOT (approved):

jadmsaadaot (changes requested):
I think this task was created to review the files under the common folder and remove what's not common. Like the Modals folder has Feedback and Schedule. Are these common? Or they belong to one page only. As for removing the Document styles and adding them to the Document Modal, the document tree is common and could be used on any page, I'm not so sure about having css defintions inside a model class

jadmsaadaot (approved):

PR #1143, opened 2023-01-25
membership add and edit

saravanpa-aot:
*Issue #:*

*Description of changes:*


met-api/migrations/versions/cad222167ce7_memberships_added.py
```diff
@@ -0,0 +1,58 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    membership_status_codes = op.create_table('membership_status_codes',
+    sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
+    sa.Column('status_name', sa.String(length=50), nullable=True),
+    sa.Column('description', sa.String(length=50), nullable=True),
+    sa.PrimaryKeyConstraint('id')
+    )
+    op.create_table('membership',
```
jadmsaadaot, line 29: Should we be specific in naming? Like EngagementMembership rather than membership?

VineetBala-AOT (approved):

met-api/migrations/versions/cad222167ce7_memberships_added.py
saravanpa-aot: good point..i was thinking to extend this for other entities as well in the future ;
like for surveys; if there comes a requirement survey can be edited a few folks etc etc... 
so we can introduce a entity_type here  and manage..
so kept it as just Membership so that we can extend it..

met-api/src/met_api/schemas/memberships.py
```diff
@@ -0,0 +1,22 @@
...
+
+class MembershipSchema(Schema):
+    """Membership schema."""
+
+    class Meta:  # pylint: disable=too-few-public-methods
+        """Exclude unknown fields in the deserialized output."""
+
+        unknown = EXCLUDE
+
+    id = fields.Str(data_key='id')
```
jadmsaadaot, line 16: should this be fields.Int?

met-api/src/met_api/schemas/memberships.py
```diff
@@ -0,0 +1,22 @@
...
+    status = fields.Str(data_key='status')
+    created_date = fields.DateTime(data_key='created_date')
+    engagement_id = fields.Str(data_key='engagement_id')
```
jadmsaadaot, line 19: also this, fields.Int?

met-api/src/met_api/schemas/memberships.py
```diff
@@ -0,0 +1,22 @@
...
+    status = fields.Str(data_key='status')
+    user_id = fields.Str(data_key='user_id')
```
jadmsaadaot, line 21: and this, fields.Int?

met-api/src/met_api/schemas/memberships.py
```diff
@@ -0,0 +1,22 @@
...
+    class Meta:  # pylint: disable=too-few-public-methods
+        """Exclude unknown fields in the deserialized output."""
+
+        unknown = EXCLUDE
+
+    id = fields.Str(data_key='id')
+    status = fields.Str(data_key='status')
+    created_date = fields.DateTime(data_key='created_date')
+    engagement_id = fields.Str(data_key='engagement_id')
+    status = fields.Str(data_key='status')
```
jadmsaadaot, line 20: status defined twice

jadmsaadaot (approved):

met-api/migrations/versions/cad222167ce7_memberships_added.py
jadmsaadaot: absolutely, I thought this is a specific table. If it's already intended to be general then fine as is

Issue #1068, opened 2023-01-04
Standardize design styles in MET (borders, Input fields & button height)


marieanik:
As a MET user, I want to see consistent design patterns in MET to help me use the app, trust it, and make me want to keep using it.

Let's make sure all the following elements have specific styles across MET:

**Buttons** We have all the right colours but many buttons are taller than 40px. All buttons in MET must have a height of 40px.

**Input fields** We have input fields with different height. All Input fields (*not including TextArea) must have a height of 40px.

**Borders** To softer the look and make the element stand out, let's change all the borders in MET to:
border: 1px solid #cdcdcd
box-shadow: rgb(0 0 0 / 6%) 0px 2px 2px -1px, rgb(0 0 0 / 6%) 0px 1px 1px 0px, rgb(0 0 0 / 6%) 0px 1px 3px 0px;

![Screenshot 2023-01-04 at 7.06.00 AM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/e726e0ab-9ede-49ee-9ad2-0e6da90c3f2e)

jadmsaadaot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/1068?planning-poker) with Zenhub @djnunez-aot @saravanpa-aot @tiagograf-AOT @VineetBala-AOT

PR #1231, opened 2023-02-13
Update engagement form styles

djnunez-aot:
*Issue #1068 :*
https://github.com/bcgov/met-public/issues/1068

-Add softer borders to rich content editor

-Add softer borders to engagement text fields

-fix tab spacing

### DIFF
--- a/met-web/src/components/engagement/form/EngagementWidgets/Phases/PhasesOptionCard.tsx
+++ b/met-web/src/components/engagement/form/EngagementWidgets/Phases/PhasesOptionCard.tsx
@@ -16,6 +16,8 @@ export const optionCardStyle = {
     cursor: 'pointer',
     '&:hover': { backgroundColor: 'rgb(242, 242, 242)' },
     minHeight: '110px',
+    border: '1px solid #cdcdcd',
+    boxShadow: 'rgb(0 0 0 / 6 %) 0px 2px 2px - 1px, rgb(0 0 0 / 6 %) 0px 1px 1px 0px, rgb(0 0 0 / 6 %) 0px 1px 3px 0px',
 };
 
 const PhasesOptionCard = () => {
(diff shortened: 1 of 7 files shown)


met-web/src/components/engagement/form/RichTextEditor.tsx
```diff
@@ -60,7 +60,7 @@ const RichTextEditor = ({
 
     return (
         <FormControl fullWidth>
-            <MetPaper style={{ borderColor: `${error ? '#d32f2f' : '#606060'}` }}>
+            <MetPaper style={{ borderColor: `${error ? '#d32f2f' : ''}` }}>
```
jadmsaadaot, line 63: why? Don't we want this border to be red when there's error?

met-web/src/components/engagement/form/EngagementWidgets/Phases/PhasesOptionCard.tsx
```diff
@@ -16,6 +16,8 @@ export const optionCardStyle = {
     cursor: 'pointer',
     '&:hover': { backgroundColor: 'rgb(242, 242, 242)' },
     minHeight: '110px',
+    border: '1px solid #cdcdcd',
+    boxShadow: 'rgb(0 0 0 / 6 %) 0px 2px 2px - 1px, rgb(0 0 0 / 6 %) 0px 1px 1px 0px, rgb(0 0 0 / 6 %) 0px 1px 3px 0px',
```
jadmsaadaot, line 20: If we do that in StyledPaper, do we need to it here?

met-web/src/components/engagement/form/RichTextEditor.tsx
djnunez-aot: the border is still red when there is an error it is just going to inherit the other border style when there is no error

met-web/src/components/engagement/form/EngagementWidgets/Phases/PhasesOptionCard.tsx
djnunez-aot: good catch no we do not!

met-web/src/components/engagement/form/RichTextEditor.tsx
jadmsaadaot: ohhh right! Then let's something like sx={[ error && { borderColor:...}]}. This way borderColor is not set to a bad value like '' when there's no error
image

jadmsaadaot (approved):

VineetBala-AOT (approved):

PR #1254, opened 2023-02-16
Update spacing for mobile responsiveness

djnunez-aot:
-Add spacing for the toolbar

### DIFF
--- a/met-web/src/components/common/index.tsx
+++ b/met-web/src/components/common/index.tsx
@@ -18,7 +18,7 @@ import { MET_Header_Font_Family, MET_Header_Font_Weight } from '../../styles/con
 import { When } from 'react-if';
 
 export const MobileToolbar = styled(Toolbar)(() => ({
-    paddingBottom: 10,
+    marginBottom: '40px',
 }));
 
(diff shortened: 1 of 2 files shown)


met-web/src/components/common/index.tsx
```diff
@@ -16,6 +17,10 @@ import LoadingButton from '@mui/lab/LoadingButton';
 import { MET_Header_Font_Family, MET_Header_Font_Weight } from '../../styles/constants';
 import { When } from 'react-if';
 
+export const MobileToolbar = styled(Toolbar)(() => ({
```
jadmsaadaot, line 20: I forgot why we're using a toolbar for spacing instead of just adding some margin. This a bit confusing to any maintainer. Could we find a better way to have that spacing. A <span> with padding would at least make more sense if there isn't any other idea

jadmsaadaot (changes requested):
Better way to have spacing than adding an empty component called toolbar

jadmsaadaot (approved):

saravanpa-aot (approved):

PR #1271, opened 2023-02-24
Add edit & delete functions to event card

djnunez-aot:
*Issue #:*

-Work in progress


met-web/src/components/engagement/form/EngagementWidgets/Events/VirtualSessionFormDrawer.tsx
```diff
@@ -31,29 +36,65 @@ type VirtualSessionForm = yup.TypeOf<typeof schema>;
...
+    }, [eventToEdit]);
+
     const { handleSubmit, reset } = methods;
 
-    const onSubmit: SubmitHandler<VirtualSessionForm> = async (data: VirtualSessionForm) => {
-        if (!widget) {
-            return;
+    const updateEvent = async (data: VirtualSessionForm) => {
+        if (eventToEdit && widget) {
+            console.log(eventToEdit);
```
jadmsaadaot, line 78: console log

met-api/src/met_api/models/event_item.py
```diff
@@ -29,3 +31,36 @@ class EventItem(BaseModel):  # pylint: disable=too-few-public-methods, too-many-
     def save_event_items(cls, event_items: list) -> None:
         """Update widgets.."""
         db.session.bulk_save_objects(event_items)
+        
+    
+    @classmethod
+    def update_event_item(cls, event_data: dict) -> Optional[EventItem or DefaultMethodResult]:
+        """Update Event Item."""
+        query = EventItem.query.filter_by(id=event_data.get('id'))
+        update_fields = dict(
```
jadmsaadaot, line 40: If we're patching we don't want to ideally touch all fields when we update, we can pass only the affected fields. We can follow the pattern in the engagement model. edit_engagment is what we did so we don't do it the update_engagement way and send all fields to the update
image

met-api/src/met_api/models/event_item.py
```diff
@@ -29,3 +31,36 @@ class EventItem(BaseModel):  # pylint: disable=too-few-public-methods, too-many-
...
+            start_date=event_data.get('start_date', None),
+            end_date=event_data.get('end_date', None),
+            url=event_data.get('url', None),
+            url_label=event_data.get('url_label', None),
+            sort_index=event_data.get('sort_index', None),
+            widget_events_id=event_data.get('widget_events_id', None),
+            updated_date=datetime.utcnow(),
+        )
+        if not query:
+            return DefaultMethodResult(False, 'Event Not Found', event_data.get('widgets_event_id'))
```
jadmsaadaot, line 53: I think we have moved away from DefaultMethodResult. If it's still used in the codebase flag it in the group so we create a task to remove them

met-api/src/met_api/models/event_item.py
djnunez-aot: good idea there

met-api/src/met_api/models/event_item.py
djnunez-aot: where would you like me to flag it

jadmsaadaot: If it's a lot, then we can just create a task to find and remove them, maybe there's one already?

jadmsaadaot (approved):
Good job!

Issue #1285, opened 2023-02-27
Change line-spacing on buttons to 1.1rem


marieanik:
![Screenshot 2023-02-27 at 2.10.10 PM.png](https://images.zenhubusercontent.com/63c595f1d8ede2145e2c4ad4/a2c0135c-fb92-4b5b-af75-031a62b80f25)Sometimes, resizing the screen forces the text to wrap. The current line spacing is set too big so the text ends up spilling outside the border of the button. Changing the line-spacing to 1.1 will solve this issue

![Screenshot 2023-02-27 at 12.21.51 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/9541d46d-2dec-459e-833a-4453ba358e41)![Screenshot 2023-02-27 at 12.22.06 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/f60d8044-ce5b-4cd8-8b97-71b9fab0a4e6)ue.

MA note: should be solved by https://app.zenhub.com/workspaces/met-ops-6254279dcfa60b0010ea5a0d/issues/gh/bcgov/met-public/1285

Button text overlaps container: Either decrease the text size or shorten the phrase- maybe just 'sign up'
Screenshot 2023-02-27 at 2.10.10 PM.png

PR #1319, opened 2023-03-03
Change button lineheight to 1.1rem

djnunez-aot:
*Issue #1285 :*
https://github.com/bcgov/met-public/issues/1285

-Update button line height

### DIFF
--- a/met-web/src/styles/Theme.ts
+++ b/met-web/src/styles/Theme.ts
@@ -62,6 +62,7 @@ export const BaseTheme = createTheme({
         MuiButton: {
             styleOverrides: {
                 root: {
+                    lineHeight: '1.1rem',
                     height: '40px',
                 },
             },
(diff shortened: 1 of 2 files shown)


met-web/src/styles/Theme.ts
```diff
@@ -62,6 +62,7 @@ export const BaseTheme = createTheme({
         MuiButton: {
             styleOverrides: {
                 root: {
+                    lineHeight: '1.1rem',
```
jadmsaadaot, line 65: can we remove the style overrides here and just add them to PrimaryButton and SecondaryButton

jadmsaadaot (changes requested):
Move overrides to common atomic components

met-web/src/styles/Theme.ts
djnunez-aot: sure thing

jadmsaadaot: have we also removed any other style overrides here and moved them to the atomic components?

saravanpa-aot (approved):

Issue #1290, opened 2023-02-28
Add Lat and Long in the map widget


saravanpa-aot:
As an admin user/team member,
I should be able to add a lat long to engagement description 
so that I can show a map on an engagement page.

Matches wireframe: https://xd.adobe.com/view/61e16dd6-919e-4374-a074-d09fd2184766-34b2/screen/b78255eb-d9d8-4c6f-9b7d-4345886afcbc/specs/

**ACCEPTANCE CRITERIA**

- Need a table in the backend
- Form in the front end
- Map preview as back end user puts in lat/long
- Matches wireframe

COULD add a "preview" to open up modal of what the map looks like, but let's explore as we design

Alternate "preview" version:

https://xd.adobe.com/view/61e16dd6-919e-4374-a074-d09fd2184766-34b2/screen/3aab1a4f-2575-4915-a274-3eaa19b3425c/specs/

https://xd.adobe.com/view/61e16dd6-919e-4374-a074-d09fd2184766-34b2/screen/17c72644-856f-4645-8e8e-9c8026fbc326/specs/

carlygreenehill:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/1290?planning-poker) with Zenhub @djnunez-aot @Ian-UX-AOT @jadmsaadaot @marieanik @saravanpa-aot @VineetBala-AOT

carlygreenehill:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/1290?planning-poker) with Zenhub @Connor1892

PR #1324, opened 2023-03-06
Map widget model 

djnunez-aot:
*Issue #1290:*
https://github.com/bcgov/met-public/issues/1290

-Add map widget model

-Add map widget schema

-Add map widget resource

-Add map widget service


met-api/src/met_api/models/map.py
```diff
@@ -0,0 +1,59 @@
...
+
+class WidgetMap(BaseModel):  # pylint: disable=too-few-public-methods
+    """Definition of the Map entity."""
+
+    __tablename__ = 'map'
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    widget_id = db.Column(db.Integer, ForeignKey('widget.id', ondelete='CASCADE'), nullable=True)
+    engagement_id = db.Column(db.Integer, ForeignKey('engagement.id', ondelete='CASCADE'), nullable=True)
+    title = db.Column(db.String(50))
+    latitutde = db.Column(db.Float)
```
jadmsaadaot, line 22: lat and long should be nullable=False I think

djnunez-aot: what do we display if we have a map without longitude and latitude is that if it's just a shapefile

jadmsaadaot: We are not uploading shapefiles now, so let's not worry about that. And we shouldn't at the moment create these widgets without lats and longs

met-api/src/met_api/models/map.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+from __future__ import annotations
+
+from sqlalchemy.sql.schema import ForeignKey
+
+from .base_model import BaseModel
+from .db import db
+
+
+class WidgetMap(BaseModel):  # pylint: disable=too-few-public-methods
```
jadmsaadaot, line 13: please also change the file name map.py to something like widget_map. I'm thinking map is a very general term that could apply to any potential map component that isn't related to widgets

djnunez-aot: done

met-api/migrations/versions/a1237c8a3df9_.py
```diff
@@ -0,0 +1,46 @@
...
+    sa.Column('description', sa.String(length=500), nullable=True),
+    sa.Column('longitude', sa.Float(), nullable=False),
+    sa.Column('latitude', sa.Float(), nullable=False),
+    sa.Column('engagement_id', sa.Integer(), nullable=True),
+    sa.Column('widget_id', sa.Integer(), nullable=True),
+    sa.Column('created_date', sa.DateTime(), nullable=False),
+    sa.Column('updated_date', sa.DateTime(), nullable=True),
+    sa.Column('created_by', sa.String(length=50), nullable=True),
+    sa.Column('updated_by', sa.String(length=50), nullable=True),
+    sa.ForeignKeyConstraint(['widget_id'], ['widget.id'], ondelete='CASCADE'),
```
tiagograf-AOT, line 33: Not sure i unnderstand this foreign key structure, If I remember correctly our Widget and Widget item tables contains a widget_data_id and we usually use that column to provide. the item id (in this case the widgetmap.id), did something change?

jadmsaadaot: Yeah I think we haven't followed that same pattern with documents and events. Widget_item could be refactored to like widget_contact. Instead of using widget_item with its data id, we created widget_events or widget_documents and just had the widge_id there.

saravanpa-aot (approved):

jadmsaadaot (approved):

Issue #1370, opened 2023-03-15
"Empty" comments shouldn't require review, and show up in the "read comment" (internal and external)


marieanik:
If no comments found in a submission it should be automatically approved.
Filter out submissions that have no comments in the listing pages (list and read all).

saravanpa-aot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/1370?planning-poker) with Zenhub @djnunez-aot @jadmsaadaot

saravanpa-aot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/1370?planning-poker) with Zenhub @tiagograf-AOT @VineetBala-AOT

tiagograf-AOT:
If no comments found in a submission it should be automatically approved.
Filter out submissions that have no comments in the listing pages (list and read all).

PR #1388, opened 2023-03-20
adding logic to auto aprv submission without cmnt

VineetBala-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/1370

*Description of changes:*
changes to auto approve submissions which:
- do not have a textarea or textfield in the submission json
- has textarea or textfield in the submission json but have a null value

changes to exclude auto approved submissions from comment listing page as they do not have a comment

### DIFF
--- a/met-api/src/met_api/models/submission.py
+++ b/met-api/src/met_api/models/submission.py
@@ -5,7 +5,7 @@ Manages the Submission
 from __future__ import annotations
 from datetime import datetime
 from typing import List
-from sqlalchemy import TEXT, ForeignKey, asc, cast, desc, or_, text
+from sqlalchemy import TEXT, ForeignKey, and_, asc, cast, desc, or_, text
 from sqlalchemy.dialects import postgresql
 
 from met_api.constants.comment_status import Status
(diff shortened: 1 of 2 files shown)


jadmsaadaot (approved):

djnunez-aot (approved):

met-api/src/met_api/models/submission.py
```diff
@@ -53,6 +53,16 @@ def get_by_survey_id(cls, survey_id) -> List[SubmissionSchema]:
     @classmethod
     def create(cls, submission: SubmissionSchema, session=None) -> Submission:
         """Save submission."""
+        has_comments = cls.__check_if_submission_has_comments(submission)
```
saravanpa-aot, line 56: can you add some comments why we are setting to approved if no comments?
just to explain the logic..

saravanpa-aot (commented):
Can you add a test for this?
Do a sumbission ,
Assert the status is approved

met-api/src/met_api/models/submission.py
```diff
@@ -72,6 +84,31 @@ def create(cls, submission: SubmissionSchema, session=None) -> Submission:
             session.flush()
         return new_submission
 
+    @staticmethod
```
saravanpa-aot, line 87:
```
@staticmethod
def __check_if_submission_has_comments(submission: SubmissionSchema):
    """Check if comment exists."""
    submission_json = submission.get('submission_json', {})
    text_fields = ['simpletextarea', 'simpletextfield']

    for field in text_fields:
        if field in submission_json and len(submission_json[field]) > 0:
            return True

    return False
```

saravanpa-aot: can we simply this?
do check the logic in detail please..i am not sure 100% if it works

met-api/src/met_api/models/submission.py
VineetBala-AOT: added the comments, thanks

met-api/src/met_api/models/submission.py
VineetBala-AOT: Changed the logic as suggested, thanks.

saravanpa-aot (approved):
great job!!

Issue #1006, opened 2022-12-14
Add mutation functions for Widget in Redux RTK apis


jadmsaadaot:

carlygreenehill:
Not 3 points! Need to re-estimate

PR #1391, opened 2023-03-21
Create widget service mutation functions in RTK

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/1006

*Description of changes:*
- add createWidget Mutation
- add deleteWidget mutation
- add sortWidgets mutation
- invalidate cache accordingly when above mutations are called
- update unit tests
- fix unit test warning regarding using xs prop for Grid that isn't item


djnunez-aot (approved):

met-web/src/components/engagement/form/EngagementWidgets/Map/MapOptionCard.tsx
```diff
@@ -52,23 +54,16 @@ const MapOptionCard = () => {
...
                 <Else>
-                    <Grid
-                        xs={12}
-                        container
-                        alignItems="center"
-                        justifyContent="flex-start"
-                        direction="row"
-                        columnSpacing={1}
-                    >
+                    <Grid container alignItems="center" justifyContent="flex-start" direction="row" columnSpacing={1}>
```
saravanpa-aot, line 66: these changes are intended rite?
xs={12} got removed..

jadmsaadaot: yes, it was causing an error in jest. And it was not effective anyway. xs prop only works if grid is item. Above wasn't item so xs had no effect and just caused warnings

Issue #1384, opened 2023-03-20
Comments on engagements page (UI)


marieanik:
As a MET system user, I want to be able to access the comments from the engagement and survey listing pages and see clearly how many comments there are for each status.

Viewers: See only "Approved Comments badge" > All already approved comments
![Screenshot 2023-03-20 at 1.52.03 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/36e21f24-eff5-41aa-b766-e280e444c6ec)
Team Members: See only "Approved Comments badge" > All already approved comments
                           See comments badges & View All for projects they are assigned to.
![Screenshot 2023-03-20 at 1.52.36 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/41ceda4f-31f9-4142-ae7d-b73da88442e8)

Superusers: See comments badges & "View All" for all projects.
![Screenshot 2023-03-20 at 1.54.49 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/9a6717c9-8e6a-474e-8629-a8e39cdaa0d4)

PR #1409, opened 2023-03-22
Add comment status icons to engagement listing table

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/1384

*Description of changes:*

- extend get_submissions_meta_data in the Engagement schema to add approved, pending, rejected and new submissions count
- Create icons for approved, pending, rejected and new
- Add icons to table
- add few helper props to HeadCells interface


jadmsaadaot:
I notice 4 code smells were flagged, I'm going to ignore them for now. It flagged getValue in HeadCells. getValue is something we are already doing for our HeadCells array

djnunez-aot (approved):

met-api/src/met_api/schemas/engagement.py
```diff
@@ -53,6 +54,14 @@ def get_submissions_meta_data(self, obj):
             }
         return {
             'total': len(obj.surveys[0].submissions),
+            'pending': len([submission for submission in obj.surveys[0].submissions
```
saravanpa-aot, line 57: Can we make this into a method by any chance?
a method to return the count if status is passed that way its more readable

`counts['pending'] = count_submissions_by_status(survey.submissions, CommentStatus.Pending)`
the method just does same thing in the above code

```
def count_submissions_by_status(submissions, status):
        return len([s for s in submissions if s.comment_status_id == status.value])
    
```

but since its already in return statement , i am not too sure how easy will be this.. but if we have a way to reduce the duplication , its great

jadmsaadaot: Done

saravanpa-aot (approved):
please add tests ;
well done

jadmsaadaot:
> please add tests ; well done

Will merge now, and add tests in follow up PR. To prevent conflicts.

Issue #1382, opened 2023-03-16
Preview/View show upload shapefile  


saravanpa-aot:
UX Notes:

If a shapefile is uploaded, Lat/Long should be optional (we can leave it mandatory until we get to it)
I saw a bug was created for the map not showing so I'll ignore that for this task. 
Can you please add the description for the Lat, Long. and marker labels? (Same style as on the engagement details page. You can see them on wireframes.)

As a MET system user, I want to be able to select, upload a shape file so that the polygon can be displayed on the map widget.

### Scenario 1: 

User accesses the map widget drawer, clicks a button to upload a shape file. File uploads and card appears with the name of the file and the option to remove the file (similar to other widgets). 

Users clicks "Preview Map" button and sees a preview of the shapefile on the map.

### Scenario 2:

Matches wireframes:
https://xd.adobe.com/view/314be723-38ea-4f16-acb4-69db36ac71f9-6418/screen/3aab1a4f-2575-4915-a274-3eaa19b3425c/
https://xd.adobe.com/view/314be723-38ea-4f16-acb4-69db36ac71f9-6418/screen/026876b8-374a-4463-bacd-6d7386e401a8/

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

djnunez-aot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/1382?planning-poker) with Zenhub @jadmsaadaot @saravanpa-aot @tiagograf-AOT @VineetBala-AOT

VineetBala-AOT:
Option to remove the file not showing after the widget is saved.

PR #1423, opened 2023-03-24
touch ups to shapefile form

djnunez-aot:
*Issue #1382 :*
https://github.com/bcgov/met-public/issues/1382

-Add file name to widget_map

-display if any files have been previously updated

-change preview button location

-add unit tests for shapefile form


met-api/src/met_api/services/widget_map_service.py
```diff
@@ -36,13 +36,15 @@ def update_map(widget_id, request_json):
...
     @staticmethod
-    def _create_map_model(widget_id, map_data: dict, geojson=None):
+    def _create_map_model(widget_id, map_data: dict, geojson=None, filename=None):
         map_model: WidgetMapModel = WidgetMapModel()
         map_model.widget_id = widget_id
         map_model.marker_label = map_data.get('marker_label')
         map_model.latitude = map_data.get('latitude')
         map_model.longitude = map_data.get('longitude')
         map_model.engagement_id = map_data.get('engagement_id')
+        if filename:
```
jadmsaadaot, line 46: is it awkward to have the filename and geojson inside the map_data dict?

met-api/migrations/versions/141eb11bdffc_merge_heads.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '141eb11bdffc'
+down_revision = ('e3427bc518f5', 'f99eb7f53041')
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    pass
```
tiagograf-AOT, line 20: empty migration shouldn't ever happen

djnunez-aot: Yes theres different head files with the other prs

met-api/src/met_api/services/widget_map_service.py
jadmsaadaot: As part of the second arguments passed the map_data

met-web/src/components/engagement/form/EngagementWidgets/Map/PreviewModal.tsx
```diff
@@ -29,8 +29,8 @@ export const PreviewModal = () => {
                 >
                     <Map
                         geojson={previewMap.geojson}
-                        longitude={previewMap.longitude}
-                        latitude={previewMap.latitude}
+                        longitude={previewMap.longitude ? previewMap.longitude : undefined}
```
jadmsaadaot, line 32: Do we need to explicitly pass undefined? when previewMap.longitude fails this ternary it would be already undefined no?

met-web/src/components/engagement/form/EngagementWidgets/Map/Form.tsx
```diff
@@ -114,8 +146,8 @@ const Form = () => {
             });
         }
         setPreviewMap({
-            longitude: validatedData.longitude,
-            latitude: validatedData.latitude,
+            longitude: validatedData.longitude ? validatedData.longitude : undefined,
+            latitude: validatedData.latitude ? validatedData.latitude : undefined,
```
jadmsaadaot, line 150: same  comment about undefined

met-web/src/components/map/index.tsx
```diff
@@ -10,8 +10,8 @@ import { When } from 'react-if';
 import { AnyLayer } from 'mapbox-gl';
 
 interface MapProps {
-    latitude: number;
-    longitude: number;
+    latitude?: number;
+    longitude?: number;
```
jadmsaadaot, line 14: what happens when we have lat and long undefined and pass them like that to the map?

met-web/src/components/engagement/form/EngagementWidgets/Map/Form.tsx
```diff
@@ -14,23 +21,48 @@ import { WidgetDrawerContext } from '../WidgetDrawerContext';
...
+                then: yup
+                    .number()
+                    .nullable()
+                    .min(-180, 'Longitude must be greater than or equal to -180')
+                    .max(180, 'Longitude must be less than or equal to 180'),
+                otherwise: yup
+                    .number()
+                    .nullable()
+                    .min(-180, 'Longitude must be greater than or equal to -180')
+                    .max(180, 'Longitude must be less than or equal to 180'),
```
jadmsaadaot, line 64: what is the difference between then and otherwise in this case?

jadmsaadaot:
In my opinion, I think we should keep lat and long required. The map widget should always receive lat and long and never passed undefined. As for the user, I still would prefer that they always input lat and long. Shapefile is a layer, I don't think it should be treated as coordinates. What if multiple shapefiles were added, from which do we take the coordinates? Even if they contain lat and long, the admin should be responsible for extracting and inputting them. There should be an independently inputted focus area. What do you think @saravanpa-aot @tiagograf-AOT @djnunez-aot

djnunez-aot (commented):
@jadmsaadaot I agree as well i think lat & long would be better mandatory.

tiagograf-AOT (approved):

Issue #1389, opened 2023-03-20
BUG Cannot enter a comment in Rate My App - in Edge Browser


carlygreenehill:
![image](https://user-images.githubusercontent.com/101662408/226483181-50e885ed-9e65-443c-984f-875708fd83a2.png)

Can't hit submit after writing in text for Rate My App - button is greyed out both in front end and back end. 

Ma note: I experience the same issue on a MacBook pro in Chrome.

tiagograf-AOT:
Fix confirmed

PR #1425, opened 2023-03-24
fix for feedback comment without rating

VineetBala-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/1389

*Description of changes:*
- Changes made to add a new type for rating i.e. None. In case the user does not provide a rating the rating type we be considered as None.
- Enable user to submit a comment with comment type but without selecting a rating


met-web/src/components/feedback/FeedbackModal/index.tsx
```diff
@@ -232,7 +232,10 @@ export const FeedbackModal = () => {
                                 <PrimaryButton
                                     data-testid="submit-button"
                                     loading={isSaving}
-                                    disabled={Boolean(!rating || (comment_type !== CommentTypeEnum.None && !comment))}
+                                    disabled={Boolean(
+                                        (rating === RatingTypeEnum.None && comment_type === CommentTypeEnum.None) ||
```
jadmsaadaot, line 236: can we extract these two into variables with meaningful names then use the variables here instead.

VineetBala-AOT: updated to have the conditions in a variable. thanks.

jadmsaadaot (approved):

Issue #1453, opened 2023-03-30
"Multi-pages" option location/disclaimer 


carlygreenehill:
**As a** Superuser,

**I want** to be prompted to make a multi-page survey vs single page survey

**so that** (reason/ why/ value) I know how to begin designing my survey without losing any unsaved work.

### Scenario 1: 

GIVEN ... a user hits "create new survey"
WHEN ... they see a prompt/toggle which gets them to choose multi or single
THEN ... they'll see what type it is before they start inputting questions, with reminder that is they choose to change half way through they will lose their work

![image](https://user-images.githubusercontent.com/101662408/228973118-e6ec504b-5bc3-4c03-b948-c96d568d0355.png)

### Scenario 2:
 xxxxxx

### Validation Rules? 

(If yes, list here)

### Screenshot 

(Link above will always be the most up-to-date version.)

VineetBala-AOT:
Validated, user has option to select single/multi page when they hit "create new survey"

PR #1456, opened 2023-04-03
Add multi page option in survey creation

djnunez-aot:
*Issue #1453 :*
https://github.com/bcgov/met-public/issues/1453

-Added auto-complete component to start survey as multi-page

image

-Added a modal to warn users that switching survey type will remove all their current progress

image

### DIFF
--- a/met-web/src/components/survey/building/index.tsx
+++ b/met-web/src/components/survey/building/index.tsx
@@ -32,7 +32,7 @@ const SurveyFormBuilder = () => {
     const [savedEngagement, setSavedEngagement] = useState<Engagement | null>(null);
 
     const [formDefinition, setFormDefinition] = useState<FormBuilderData>({ display: 'form', components: [] });
-
+    const isMultiPage = formDefinition.display === 'wizard';
     const hasEngagement = Boolean(savedSurvey?.engagement_id);
     const isEngagementDraft = savedEngagement?.status_id === EngagementStatus.Draft;
     const hasPublishedEngagement = hasEngagement && !isEngagementDraft;
@@ -228,13 +228,9 @@ const SurveyFormBuilder = () => {
                                                 subText: [
                                                     {
                                                         text: `You will be changing the survey type from ${
-                                                            formDefinition.display === 'wizard'
-                                                                ? 'multi page'
-                                                                : 'single page'
-                                                        } to ${
-                                                            formDefinition.display === 'wizard'
-                                                                ? 'single page'
-                                                                : 'multi page'
+                                                            isMultiPage
+                                                                ? 'multi page to single page'
+                                                                : 'single page to multi page'
                                                         }.`,
                                                     },
                                                     {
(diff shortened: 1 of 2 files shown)


met-web/src/components/survey/building/index.tsx
```diff
@@ -219,11 +220,38 @@ const SurveyFormBuilder = () => {
...
+                                    dispatch(
+                                        openNotificationModal({
+                                            open: true,
+                                            data: {
+                                                header: 'Change Survey Type',
+                                                subText: [
+                                                    {
+                                                        text: `You will be changing the survey type from ${
+                                                            formDefinition.display === 'wizard' ? 'multi-page' : 'form'
+                                                        } to ${
```
saravanpa-aot, line 232: is there a way not to repeat this same check in three places?
formDefinition.display === 'wizard' ? 'multi-page' : 'form'

met-web/src/components/survey/create/CreateOptions.tsx
```diff
@@ -94,6 +100,30 @@ export const CreateOptions = () => {
                     helperText={getErrorMessage()}
                 />
             </Grid>
+            <Grid item xs={6}></Grid>
+            <Grid item xs={6}>
+                <MetLabel sx={{ marginBottom: '2px' }}>Select Survey Type</MetLabel>
+                <Autocomplete
+                    id="survey-type-selector"
+                    options={[SurveyType.Single, SurveyType.Multi]}
```
jadmsaadaot, line 108: Autocomplete is overkill for this. We already know it's either single or multi. Toggle doesn't work? or if drop down is needed, I still prefer toggle, a simple textfield with select prop should work no?

met-web/src/components/survey/building/index.tsx
```diff
@@ -219,11 +220,38 @@ const SurveyFormBuilder = () => {
...
+                                                            formDefinition.display === 'wizard' ? 'form' : 'multi-page'
```
jadmsaadaot, line 233: from form to multi-page. there both forms right. Maybe from single page to multi page

met-web/src/components/survey/building/index.tsx
```diff
@@ -219,11 +220,42 @@ const SurveyFormBuilder = () => {
...
+                                            data: {
+                                                header: 'Change Survey Type',
+                                                subText: [
+                                                    {
+                                                        text: `You will be changing the survey type from ${
+                                                            formDefinition.display === 'wizard'
+                                                                ? 'multi page'
+                                                                : 'single page'
+                                                        } to ${
+                                                            formDefinition.display === 'wizard'
```
tiagograf-AOT, line 235: both conditions will always have the same value, just use one condition and full text instead of checking it twice.

Also, make this a boolean flag instead

tiagograf-AOT (approved):

jadmsaadaot (approved):

Issue #1418, opened 2023-03-23
Empty comments should be counted in the "Total" comment count on survey listing page


marieanik:
The number of responses on the table shows the total including the "empty' comments. It should only display the total number of "text" comments.

***Might be solved with the "comment badges": https://app.zenhub.com/workspaces/met-ops-6254279dcfa60b0010ea5a0d/issues/gh/bcgov/met-public/1409

PR #1460, opened 2023-04-03
Add comment count badges to survey listing page

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/1418

*Description of changes:*
- Add comment count badges to survey listing page
image

### DIFF
--- a/met-web/src/components/survey/listing/index.tsx
+++ b/met-web/src/components/survey/listing/index.tsx
@@ -11,9 +11,15 @@ import TextField from '@mui/material/TextField';
 import SearchIcon from '@mui/icons-material/Search';
 import Stack from '@mui/material/Stack';
 import { getSurveysPage } from 'services/surveyService';
-import { useAppDispatch } from 'hooks';
+import { useAppDispatch, useAppSelector } from 'hooks';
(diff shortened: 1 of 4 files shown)


met-web/src/components/survey/listing/index.tsx
```diff
@@ -155,22 +165,112 @@ const SurveyListing = () => {
...
+                        View All
                     </MuiLink>
                 );
             },
         },
+        {
+            key: 'comments_meta_data',
+            numeric: true,
+            disablePadding: false,
+            label: '',
```
saravanpa-aot, line 190: could this be duplicate code of engagement page?

tiagograf-AOT (approved):

met-web/src/components/survey/listing/index.tsx
jadmsaadaot: It's the same columns in the UI, with few differences in the key and the object they index, here it's row.comments_meta_data, engagement page is row.submissions_meta_data. Here the link is row.id, there it's row.surveys[0].id. So minor changes that might make it awkward to extract them into a common place.

saravanpa-aot (approved):

Issue #1366, opened 2023-03-15
Saving a survey as Draft or Approved


marieanik:
As a superuser, I want to flag my survey as "Hidden" while I work on it so it will not be visible to anyone until it is ready.

Superuser is on the "form.io Create Survey" page, and toggle on the "Hide Survey" toggle and save the Survey. Survey will only be visible to Superuser as long as the toggle is on.

This option will be disabled for TM.

Tooltip: When you toggle ON this option and save your Survey, your Survey will be "Hidden". As long as this option is on, the Survey will only be visible to Superusers. When you are ready to make it available, change the toggle to OFF and click the Save button.

matches wireframe: https://xd.adobe.com/view/314be723-38ea-4f16-acb4-69db36ac71f9-6418/screen/cefb6c98-337c-426b-96c6-e45dc617ee62/

Spreadsheet with all the options: `https://bcgov.sharepoint.com/:x:/r/teams/04610/Shared%20Documents/General/User%20Management.xlsx?d=w11f97ab36fdd489aaaa26bb5a35daa87&csf=1&web=1&e=SPkc20`

tiagograf-AOT:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/1366?planning-poker) with Zenhub @djnunez-aot @jadmsaadaot @saravanpa-aot @VineetBala-AOT

tiagograf-AOT:
New IsDraft column
"New" endpoint to get only "approved" surveys
Only "approved" surveys can be linked to engagement
Disable create survey button if not "Admin"

PR #1470, opened 2023-04-06
saving a survey as hidden

VineetBala-AOT:
*Issue #:*
https://github.com/bcgov/met-public/issues/1366

*Description of changes:*
- Logic to add a hide survey toggle while saving a survey. Default value to be false unless specified by the user (admin only)
- Validation to check if set to true then survey to be visible only to admins
- Logic to exclude the hidden surveys from clone and add survey to engagement list
- Logic to disable create survey and hide survey toggle for users not having create survey access


met-api/migrations/versions/e5d3bbb8d2f0_add_survey_hidden_column.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'e5d3bbb8d2f0'
+down_revision = '7d8897c412de'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('survey', sa.Column('is_hidden', sa.Boolean(), nullable=True))
```
jadmsaadaot, line 21: I suggest to do something like this. Add the column as nullabe. Update all the null values to False then make it nullable False and default to False
image

met-api/src/met_api/models/survey.py
```diff
@@ -45,10 +46,21 @@ def get_open(cls, survey_id) -> Survey:
...
-    def get_surveys_paginated(cls, pagination_options: PaginationOptions, search_text='', unlinked=False):
+    def get_surveys_paginated(  # pylint: disable=too-many-arguments
+        cls, pagination_options: PaginationOptions, search_text='', unlinked=False, is_admin=False, exclude_hidden=False
+    ):
         """Get surveys paginated."""
         query = db.session.query(Survey).join(Engagement, isouter=True).join(EngagementStatus, isouter=True)
 
+        null_value = None
+        # exclude hidden surveys from being fetched by users other than admins
+        if is_admin:
```
jadmsaadaot, line 57: extract this section into its own function

jadmsaadaot: and perhaps go for if 
is_admin or exclude_hidden: 
filter

met-web/src/components/survey/building/index.tsx
```diff
@@ -261,6 +278,34 @@ const SurveyFormBuilder = () => {
             <Grid item xs={12}>
                 <FormBuilder handleFormChange={handleFormChange} savedForm={formDefinition} />
             </Grid>
+            <PermissionsGate scopes={[SCOPES.createSurvey]} errorProps={{ disabled: true }}>
```
jadmsaadaot, line 281: I think the PermissionGate will pass the disabled prop to its immediate child in this case Grid. But that won't have an effect. Maybe bring the PermissionGate to wrap the Switch instead

met-api/src/met_api/services/survey_service.py
```diff
@@ -33,12 +42,23 @@ def get_open(cls, survey_id):
...
+    def get_surveys_paginated(pagination_options: PaginationOptions, search_text='', unlinked=False, user_id='',
+                              exclude_hidden=False):
         """Get engagements paginated."""
+        # check if user is an admin
+        user: UserModel = UserModel.get_user_by_external_id(user_id)
+        if not user:
+            raise BusinessException(
+                error='Invalid User.',
+                status_code=HTTPStatus.BAD_REQUEST)
+        is_admin = KeycloakGroupName.EAO_IT_ADMIN.value in SurveyService._get_user_group_names(user)
```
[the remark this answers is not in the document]
VineetBala-AOT: updated to have a role. Thanks.

met-api/src/met_api/models/survey.py
VineetBala-AOT: moved this logic to the service instead of the model

met-api/src/met_api/services/survey_service.py
```diff
@@ -47,6 +58,14 @@ def get_surveys_paginated(pagination_options: PaginationOptions, search_text='',
             'total': total
         }
 
+    @staticmethod
+    def _can_user_access_hidden_surveys(user_roles):
```
tiagograf-AOT, line 62: can we invert the true/false logic here?

Theres a bit of text interpretation that doesn't make sense, 
Can user access hidden > returns TRUE means he cannot access it 
Reading the method name I'd assume that true means he can access it

tiagograf-AOT (approved):

jadmsaadaot (approved):

Issue #1464, opened 2023-04-04
Comment Page will log question answers


tiagograf-AOT:
When viewing a comment "Comment details page" only comments are loaded in the UI, but If network tab is opened, we can see the response JSON from the get comment endpoint that shows the. whole submission info:

![image.png](https://images.zenhubusercontent.com/6254616e054e9b1da214d3b1/b1adef22-8909-4b96-b278-96735d0a1c3b)

Submission info shouldn't be available to staff users.

PR #1475, opened 2023-04-06
Exclude submission json during submission fetch

jadmsaadaot:
*Issue #:*
https://github.com/bcgov/met-public/issues/1464

*Description of changes:*
- Pass exclude=['submission_json'] to submission schema during Marshalling

### DIFF
--- a/met-api/src/met_api/services/submission_service.py
+++ b/met-api/src/met_api/services/submission_service.py
@@ -36,7 +36,7 @@ class SubmissionService:
     def get(cls, submission_id):
         """Get submission by the id."""
         db_data = Submission.get(submission_id)
-        return SubmissionSchema().dump(db_data)
+        return SubmissionSchema(exclude=['submission_json']).dump(db_data)
 
     @classmethod
     def get_by_token(cls, token):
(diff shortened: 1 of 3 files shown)


saravanpa-aot (commented):
Shouldnt this be a change in the comment schema or even in the comment query not to fetch the submssion?

jadmsaadaot:
> Shouldnt this be a change in the comment schema or even in the comment query not to fetch the submssion?

We are not querying comment, we are querying the submission table in the above service. And submission_json is what we're hiding which is a column in submission. Comment schema or comment model is not touched here. The hiding has to be for specific requests because it is needed in others like when we allow users to remove PI from their comment

djnunez-aot (approved):

met-api/src/met_api/services/submission_service.py
```diff
@@ -232,7 +232,7 @@ def get_paginated(cls, survey_id, pagination_options: PaginationOptions, search_
             search_text,
         )
         return {
-            'items': SubmissionSchema(many=True).dump(items),
+            'items': SubmissionSchema(many=True, exclude=['submission_json']).dump(items),
```
tiagograf-AOT, line 235: isn't this what we use for public user editing the submission?

jadmsaadaot: this gets a page of submissions by survey id. It's used for comment listing, in both admin or public case submission json shouldn't be there.

This one below is used for public user editing the submission
image

PR #1505, opened 2023-04-14
Improve dashboard layout

jadmsaadaot:
*Description of changes:*

- Setup data fetch inside each component
- Add loading and error handling
- Make components mobile friendly
- Add common context for common data like engagement data
- Improve Grid layout
- Make it more modular


VineetBala-AOT (approved):

djnunez-aot (approved):

met-web/src/components/publicDashboard/KPI/TotalResponses.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { Grid, Skeleton } from '@mui/material';
+import SupervisorAccountIcon from '@mui/icons-material/SupervisorAccount';
+import { DASHBOARD } from '../constants';
+import { ErrorBox } from '../ErrorBox';
+
+const HEIGHT = '213px';
+const TotalResponses = () => {
+    // Sample data
+    const sampleData = [{ name: 'L1', value: 100 }];
+    const circleSize = 150;
```
saravanpa-aot, line 14: just thinking out aloud....
was just wondering if there should be some indication that its a graph in the name..
ie instead of TotalResponses ;should we call it as TotalResponsesGraph or something of that sort?

met-web/src/components/publicDashboard/Dashboard.tsx
```diff
@@ -0,0 +1,94 @@
...
+                                justifyContent="flex-end"
+                            >
+                                <PrimaryButton
+                                    data-testid="SurveyBlock/take-me-to-survey-button"
+                                    onClick={handleReadComments}
+                                >
+                                    Read Comments
+                                </PrimaryButton>
+                            </Grid>
+                            <Grid container spacing={3} item xs={12}>
```
saravanpa-aot, line 69:
```
const components = [
  <TotalResponses />,
  <CompleteResponses />,
  <ResponsesWithComment />,
  <SubmissionTrend />,
  <SurveyBar />
];

const gridItems = components.map((component, index) => (
  <Grid key={index} item xs={12} sm={4}>
    {component}
  </Grid>
));

return (
  <Grid container spacing={3}>
    {gridItems}
  </Grid>
);

```
Do you wanna consider this style? incase if we add a new component , it will be easier..

saravanpa-aot: if its not worth it , feel free to ignore it..

met-web/src/components/publicDashboard/KPI/CompleteResponses.tsx
```diff
@@ -0,0 +1,80 @@
...
+import Stack from '@mui/material/Stack';
+import { Skeleton } from '@mui/material';
+import { DASHBOARD } from '../constants';
+import { RadialBarChart, PolarAngleAxis, RadialBar } from 'recharts';
+import { MetPaper, MetHeader3 } from 'components/common';
+import { ErrorBox } from '../ErrorBox';
+
+const CompleteResponses = () => {
+    const sampleData = [{ name: 'L1', value: 80 }];
+    const [data, setData] = useState(sampleData);
```
saravanpa-aot, line 11: May be add comment to each component to say what the component does

met-web/src/components/publicDashboard/Dashboard.tsx
jadmsaadaot: I see the value that this makes it shorter, and easily extendable by adding a component in the list. I would recommend keeping the current style because it offers more clarity and control. Components don't have to have the same xs and sm and our design might change for each.  Like submission Trend and Survey bar don't have sm={4}

jadmsaadaot: When dealing with Grid Items I would rather they are separate and easily independently controlled.

met-web/src/components/publicDashboard/KPI/TotalResponses.tsx
jadmsaadaot: added Gauge to the name of these, as for the others "Trend" and "Bar" should cover it

Issue #1445, opened 2023-03-29
Support TenantId in Endpoints


saravanpa-aot:
Backend should get the tenant id in urls as a request header 

- [x] support request header tenant id
- [ ] extend check_auth to support tenant id based checks
- [x] read from header , set it to g and use in models where ever necessary (engagement model etc)

Front end can store the tenant id in either url(s) or session storage 

- [x] add tenant id to all requests. (UserService /Interceptors)
- [x] Check if we have access to tenant id across url redirects (if needed , save it in session storage)

tiagograf-AOT:
duplicate

saravanpa-aot:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MET-Ops-6254279dcfa60b0010ea5a0d/issues/bcgov/met-public/1445?planning-poker) with Zenhub @djnunez-aot @jadmsaadaot @tiagograf-AOT @VineetBala-AOT

PR #1506, opened 2023-04-17
Set Tenant Id to global and added to basemodel as well

saravanpa-aot:
https://github.com/bcgov/met-public/issues/1445

*Description of changes:*

1. Added tenant id to g
2. added saving of tenant id in basemodel
3. a cache implementation - Will work on improving it later
4. Changed survey to save using basemodel.save()


djnunez-aot (approved):

jadmsaadaot (approved):

tiagograf-AOT (approved):

tiagograf-AOT:
I like the base model approach! we'll have to think about having the tenant id in the get methods too.

VineetBala-AOT (approved):

PR #1507, opened 2023-04-17
Update comment submission to handle multiple textfield & text areas

djnunez-aot:
-Change the comment field to checker to check if a field starts with any value in the list instead

### DIFF
--- a/met-api/tests/unit/services/test_submission.py
+++ b/met-api/tests/unit/services/test_submission.py
@@ -138,6 +138,6 @@ def test_check_if_submission_can_handle_multiple_comments(session):
 
     submission = SubmissionService().create(email_verification.verification_token, submission_request)
     submission_json = submission.submission_json
-     
+
     # Assert that the function returns True since there is a comment in a text field that starts with 'simpletextfield2'
     assert 'simpletextfield2' in submission_json.keys()


saravanpa-aot (commented):
any chance we can do a test for this?
it might not be very easy to come up with a test..but i am sure it will be worth it..

tiagograf-AOT (approved):

jadmsaadaot (approved):

met-api/tests/unit/services/test_submission.py
```diff
@@ -116,3 +118,25 @@ def test_submissions_with_comment_are_not_auto_approved(session):  # pylint:disa
...
+                            'simpletextarea1': 'This is a comment 1'},
+        'survey_id': survey.id,
+        'user_id': user_details.id,
+        'verification_token': email_verification.verification_token,
+    }
+
+    submission = SubmissionService().create(email_verification.verification_token, submission_request)
+
+    # Assert that the function returns True since there is a comment in a text field that starts with 'simpletextarea'
+    assert submission is not None
```
jadmsaadaot, line 142: we could perhaps try fetching the submission and checking that the submission.comments with key simpletextfield2 exist

met-api/tests/unit/services/test_submission.py
```diff
@@ -116,3 +118,25 @@ def test_submissions_with_comment_are_not_auto_approved(session):  # pylint:disa
 
     assert submission is not None
     assert submission.comment_status_id == Status.Pending.value
+
+
+def test_check_if_submission_has_comments(session):
+    """Make sure comments in submission are approved."""
+    survey, eng = factory_survey_and_eng_model()
```
saravanpa-aot, line 125: I think this test case is for a particular scenarion rite?
In that case , does the method name and the docstring reflect it?
test_check_if_submission_has_comments is very generic
Make sure comments in submission are approved. - again vague.
It will be great if either the name or the comment imply the meaning of the test.

met-api/tests/unit/services/test_submission.py
djnunez-aot: refactored for that

saravanpa-aot (approved):

PR #1600, opened 2023-05-12
Add internal engagement flag

tiagograf-AOT:
*Description of changes:*
Add internal engagement flag and flow.


saravanpa-aot (commented):
is it easy to add a check on submission if it really came from a staff user?
like match token with idir during submission as well..

tiagograf-AOT:
> 

I am not sure, but this is a good point. 
I don't think it would be easy because we made the submission flow "public", i can take a look but we would have to allow logged in users (staff) to submit the surveys. for the scenario we are implementing the "government users" wouldn't be logging in, only entering their email.

Which this brings me to another point, it could cause some confusion that people think they have to sign in to answer it?!

met-api/src/met_api/services/email_verification_service.py
```diff
@@ -42,10 +42,18 @@ def get_active(cls, verification_token):
     def create(cls, email_verification: EmailVerificationSchema, session=None) -> EmailVerificationSchema:
         """Create an email verification."""
         cls.validate_fields(email_verification)
-        email_address = email_verification.get('email_address')
+        email_address: str = email_verification.get('email_address')
+        survey = SurveyModel.get_open(email_verification.get('survey_id'))
+        engagement: EngagementModel = EngagementModel.find_by_id(survey.engagement_id)
+        if engagement.is_internal and not email_address.endswith('@gov.bc.ca'):
```
saravanpa-aot, line 48: @gov.bc.ca ; can you move this to a constant?

saravanpa-aot (approved):

met-api/src/met_api/services/email_verification_service.py
tiagograf-AOT: done

VineetBala-AOT (approved):

djnunez-aot (approved):

PR #1645, opened 2023-05-24
Dash board calendar picker

djnunez-aot:
-added calendar picker to filter from & to date for data


met-web/package.json
```diff
@@ -14,6 +14,7 @@
         "@mui/lab": "^5.0.0-alpha.91",
         "@mui/material": "^5.6.1",
         "@mui/x-date-pickers": "^5.0.5",
+        "@mui/x-date-pickers-pro": "^6.4.0",
```
saravanpa-aot, line 17: what does the pro do?
what licence it has?

djnunez-aot: Pro is another material ui sub library, there's no licensing issue here though the issue is only when we use the DateRangeCalendar component.

djnunez-aot: the pro component is just cleaner and does not require the render input prop

saravanpa-aot: can you point to some links on the licensing? also pro vs pickers in terms of the 'pro' term?

djnunez-aot: https://mui.com/pricing/ this is the licensing chart 

the pro library picker:
`   <DatePicker
                                    value={fromDate}
                                        onChange={(newDate: Dayjs | null) => setFromDate(newDate)}
                                    />`
 The non pro picker:
 `     <DatePicker
                                    inputFormat="MM/DD/YYYY"
                              value={fromDate}
                                        onChange={(newDate: Dayjs | null) => setFromDate(newDate)}
                                    renderInput={(params) => <TextField {...params} />}
                                />`

saravanpa-aot: so are we using a non-free version? 

@jadmsaadaot thoughts here?

djnunez-aot: no this is still the free version. It's only certain components in the library that are not free.

djnunez-aot: Yea that sounds good ill revert to the regular date picker then and remove the pro library

tiagograf-AOT (approved):
It looks like same changes as the other PR?!

djnunez-aot:
The other pr is suppose to be a 1 file change. I was going to rebase once the other one got merged since they are separate tasks. I could just do put them all in one though

met-web/src/components/publicDashboard/SubmissionTrend/SubmissionTrend.tsx
```diff
@@ -46,14 +67,30 @@ const SubmissionTrend = ({ engagement, engagementIsLoading }: SubmissionTrendPro
     };
 
     useEffect(() => {
-        fetchData();
+        fetchData().catch((error) => {
+            console.error(error);
+        });
     }, [engagement.id, chartBy]);
 
+    useEffect(() => {
```
jadmsaadaot, line 75: do we have to use two useEffects to call the same function, can we do better?

jadmsaadaot: I think if we just add the dependencies [fromDate, toDate] to the first one it should work, can you please try that?

jadmsaadaot (changes requested):
Please try adding the  [fromDate, toDate] to the first useEffect and see if that works instead of using two use Effects, you wouldn't need the if (fromDate && toDate) in this case I think too

jadmsaadaot (changes requested):
Please try adding the  [fromDate, toDate] to the first useEffect and see if that works instead of using two use Effects, you wouldn't need the if (fromDate && toDate) in this case I think too

jadmsaadaot (changes requested):
Please try adding the  [fromDate, toDate] to the first useEffect and see if that works instead of using two use Effects, you wouldn't need the if (fromDate && toDate) in this case I think too

met-web/src/components/publicDashboard/SubmissionTrend/SubmissionTrend.tsx
djnunez-aot: done please review again

jadmsaadaot (approved):

PR #1669, opened 2023-05-30
Export map component to PDF

jadmsaadaot:
*Description of changes:*

- Export map component to PDF
- Clean code by extracting function to util file and adding error handling
- fix spacing
- fix display when fetch data fails
- Add backdrop when exporting pdf with percentage progress


jadmsaadaot:
I think the bugs that the sonar analysis is reporting are actually okay, it is being unnecessarily strict. The fetch data methods awaits the promise and does error handling

tiagograf-AOT (approved):

PR #1677, opened 2023-05-31
changes to manage subscription

VineetBala-AOT:
*Description of changes:*
- Added a new table to store the subscription status for a user
- Subscribe is based on the token so that it expires after the user has subscribed
- Unsubscribe option is independent of the token and based on the user id
- Changes make to email template to have the new format and variable for unsubscribe


met-api/src/met_api/resources/subscription.py
```diff
@@ -0,0 +1,87 @@
...
+            return str(err), HTTPStatus.INTERNAL_SERVER_ERROR
+
+    @staticmethod
+    @cross_origin(origins=allowedorigins())
+    def patch():
+        """Update a existing subscription partially."""
+        try:
+            requestjson = request.get_json()
+            subscription = SubscriptionSchema().load(requestjson, partial=True)
+            print(subscription)
```
jadmsaadaot, line 81: debug

met-api/src/met_api/models/subscription.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+        db.session.add(new_subscription)
+        if session is None:
+            db.session.commit()
+        else:
+            session.flush()
+        return new_subscription
+
+    @classmethod
+    def update(cls, subscription: SubscriptionSchema, session=None) -> Subscription:
```
jadmsaadaot, line 58: Is this duplicate of the below method?

met-web/src/components/engagement/view/ActionContext.tsx
```diff
@@ -43,48 +45,54 @@ export const ActionContext = createContext<EngagementViewContext>({
...
-                        subText: [
-                            {
-                                text: 'Your email has been verified. You will now receive news and updates from us.',
-                            },
-                        ],
-                    },
-                    type: 'update',
-                }),
-            );
+            if (subscriptionStatus == 'subscribe') {
```
jadmsaadaot, line 77: we should store this 'subscribe' in constant file possible under same Folder, SUBSCRIBE = 'subscribe' same for the other

met-web/src/components/engagement/view/ActionContext.tsx
```diff
@@ -43,48 +45,54 @@ export const ActionContext = createContext<EngagementViewContext>({
...
+                const token = scriptionKey;
+                const subscribed_email = await verifyEmailVerification(token);
+                const subscribed = JSON.stringify(subscribed_email);
+                await createSubscription({
+                    email_verification_id: JSON.parse(subscribed).id,
+                    user_id: JSON.parse(subscribed).user_id,
+                    is_subscribed: 'true',
+                });
+                setSubscriptionText('You have successfully confirmed your subscription. Thank you.');
```
jadmsaadaot, line 86: I think it's better if the state here was isSubscribed, just a boolean, this boolean is picked up by the Subscription component and there you decide the text

met-web/src/components/engagement/view/widgets/Subscribe/Subscribed/index.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { ActionProvider } from '../../../ActionContext';
+import Subscription from './Subscription';
+
+export const ManageSubscription = () => {
+    return (
+        <ActionProvider>
```
jadmsaadaot, line 7: I don't think we should be using a unique ActionProvider more than once. If we need state in multiple pages, we should use Redux

jadmsaadaot (changes requested):
Please look into using Redux for needing to access the subscription data in multiple pages instead of Context

met-api/src/met_api/models/subscription.py
VineetBala-AOT: removed

met-web/src/components/engagement/view/ActionContext.tsx
VineetBala-AOT: added in a constant file

saravanpa-aot (approved):

tiagograf-AOT (approved):

jadmsaadaot (approved):

PR #1689, opened 2023-06-02
Allow TMs to clone surveys

jadmsaadaot:
*Description of changes:*

- Add role clone_survey
- Assign clone_survey role to TMs and ADMIN
- create separate endpoint for cloning
- refactor how the form_json of the target survey is gotten, move that to backend


met-api/tests/unit/api/test_survey.py
```diff
@@ -28,12 +28,17 @@
 from tests.utilities.factory_utils import (
     factory_auth_header, factory_engagement_model, factory_survey_model, factory_tenant_model, set_global_tenant)
 
+surveys_url = '/api/surveys/'
```
saravanpa-aot, line 31: can you add a few more tests for clone? 
If we don't have them already

jadmsaadaot: Added a clone survey test in the api test

tiagograf-AOT (approved):

PR #1692, opened 2023-06-05
Email Queue Model and utility method

saravanpa-aot:
*Description of changes:*


VineetBala-AOT (approved):

tiagograf-AOT (approved):
We could have considered a tool to handle the queue but I like how simple this is for this scenario

PR #1695, opened 2023-06-05
Update engagement metadata

djnunez-aot:
-Add save & preview button to engagement metadata

### DIFF
--- a/met-web/src/components/engagement/form/ActionContext.tsx
+++ b/met-web/src/components/engagement/form/ActionContext.tsx
@@ -218,6 +218,13 @@ export const ActionProvider = ({ children }: { children: JSX.Element }) => {
     ): Promise<EngagementMetadata> => {
         setSaving(true);
         try {
+            if (!savedEngagement.id) {
+                dispatch(
+                    openNotification({ severity: 'error', text: 'Please save the engagement before adding metadata' }),
+                );
+                setSaving(false);
+                return engagementMetadata;
+            }
             const state = { ...engagementMetadata };
             const engagementMetadataToUpdate: EngagementMetadata = {
                 engagement_id: Number(engagementId),


met-web/src/components/engagement/form/EngagementFormTabs/EngagementSettings.tsx
```diff
@@ -229,6 +289,24 @@ const EngagementSettings = () => {
                         </Tooltip>
                     </ClickAwayListener>
                 </Grid>
+                <Grid item xs={12}>
+                    <PrimaryButton
+                        data-testid="update-engagement-button"
+                        sx={{ marginRight: 1 }}
+                        onClick={() => handleUpdateEngagement()}
```
jadmsaadaot, line 296: do we have to update all the engagement? Can we just call the update metadate at this stage

met-web/src/components/engagement/form/EngagementFormTabs/EngagementSettings.tsx
```diff
@@ -77,6 +94,49 @@ const EngagementSettings = () => {
...
+            rich_content: richContent,
+            status_block: surveyBlockList,
+        });
+
+        await handleUpdateEngagementMetadataRequest({
+            ...engagementFormData,
+            engagement_id: Number(engagementId),
+        });
+
+        navigate(`/engagements/${engagement.id}/form`);
```
jadmsaadaot, line 110: Do we have to navigate if we update the metadata

jadmsaadaot (changes requested):
Noticed we are still updating the engagement when we update the metadata, do we have to do that? Can we completely separate?

met-web/src/components/engagement/form/EngagementFormTabs/EngagementSettings.tsx
djnunez-aot: We can just update the metadata but if they go to the metadata tab after changing the engagement form tab it wont save all of their data.

jadmsaadaot: but they can still go back and press save form the form changes right?

djnunez-aot: yes but its confusing for the user i think it would be best if we save the entire form

tiagograf-AOT (approved):

met-web/src/components/engagement/form/EngagementFormTabs/EngagementSettings.tsx
jadmsaadaot: I see your point, but I see the confusion in saving the entire form when just the settings save is pressed

PR #1733, opened 2023-06-14
Changes to access user details page without a role

VineetBala-AOT:
*Description of changes:*
- Removed the check on user listing page to allow access to details page even if user is not having a group
- Added conditions on user details page to allow adding to group if use is not part of any


met-web/src/components/userManagement/userDetails/AddToEngagement.tsx
```diff
@@ -1,22 +1,40 @@
...
 import { useAppDispatch, useAppSelector } from 'hooks';
 import { debounce } from 'lodash';
 import { Engagement } from 'models/engagement';
 import axios, { AxiosError } from 'axios';
+import { Palette } from 'styles/Theme';
+import ControlledRadioGroup from 'components/common/ControlledInputComponents/ControlledRadioGroup';
 
 const schema = yup
     .object({
+        group: yup.string().nullable(),
```
jadmsaadaot, line 37: try making group required if user does not have roles. You can move the schema inside the component and do something like this:

VineetBala-AOT: updated, thanks

tiagograf-AOT (approved):

jadmsaadaot (approved):

PR #1774, opened 2023-06-26
Update public pages routes to the slug url

jadmsaadaot:
*Description of changes:*

- Add routes with the slug to replace the current ones with the ids
- Refactor the logic in public pages to fetch the engagement_id first by the slug
- Add error handling and explicit /not-found route for not found
- On get engagement slug by engagement id create one if it doesn't exist


saravanpa-aot (approved):
Lets move slug creation to alembic script
lets handle no slug silently and update will update/save

VineetBala-AOT (approved):

PR #1780, opened 2023-06-27
Handle existing engagements without slugs

jadmsaadaot:
*Description of changes:*

- Edit migration script to insert engagement slugs for existing engagements
- Edit SlugGenerationService to add get unique slugify method


VineetBala-AOT (approved):

met-api/migrations/versions/88aba309bc23_add_engagement_slug.py
```diff
@@ -30,7 +31,35 @@ def upgrade():
...
+
+    # Generate unique slugs for each engagement
+    slugify = SlugGenerationService.create_custom_unique_slugify()
+    for engagement in engagements_without_slugs:
+        engagement_id = engagement[0]
+        engagement_name = engagement[1]
+        slug = slugify(engagement_name)
+
+        # Insert the generated slug into the engagement_slug table
+        connection.execute(
```
saravanpa-aot, line 53: you can always use op.bulk_insert here.
need not change..just for awareness

saravanpa-aot (approved):

met-api/migrations/versions/88aba309bc23_add_engagement_slug.py
jadmsaadaot: thanks for noticing, replaced it with bulk insert

PR #1785, opened 2023-06-27
Update public links in app to use slug

jadmsaadaot:
*Description of changes:*

- Update the navigation links in public pages to use the slug
- Deleted the component at met-web/src/components/engagement/dashboard because it was not referenced anywhere and it is replaced by the component at met-web/src/components/publicDashboard


saravanpa-aot (commented):
the files getting deleted are not needed?
can you add a PR description to state that plz

VineetBala-AOT (approved):

PR #1798, opened 2023-07-04
Changes to handle report setting

VineetBala-AOT:
*Description of changes:*
- ETL changes to extract and load report setting to analytics database
- Analytics changes to include only approved questions on the dashboard
- Removed un-used analytics tables and models.
- Refactored ETL code
- Updated Dagster version for local setup
- Added logic to copy the report settings while a survey is cloned
- Added logic to delete report settings if question is removed from the survey form


analytics-api/src/analytics_api/models/request_type_option.py
```diff
@@ -23,25 +23,26 @@ def get_survey_result(
     ):
         """Get the analytics survey id for an engagement id."""
         analytics_survey_id = (db.session.query(SurveyModel.id)
-                               .filter(SurveyModel.engagement_id == engagement_id)
-                               .filter(SurveyModel.is_active == true())
+                               .filter_by(SurveyModel.engagement_id == engagement_id,
+                                          SurveyModel.is_active == true())
                                .subquery())
 
         # Get all the survey questions specific to a survey id which are in active status.
```
saravanpa-aot, line 30: is it postion?
not position??

saravanpa-aot (approved):
one typo probably

analytics-api/src/analytics_api/models/request_type_option.py
VineetBala-AOT: updates the column name, thanks

jadmsaadaot (approved):

PR #1810, opened 2023-07-07
Add survey report settings page

jadmsaadaot:
*Description of changes:*

- Add survey report settings page
- Add survey report services
- Add ClientSidePagination wrapper component for MetTable
- Update resource path


jadmsaadaot:
Front end unit tests will follow this PR

VineetBala-AOT (approved):

met-api/src/met_api/services/survey_service.py
```diff
@@ -151,7 +152,14 @@ def update(cls, data: SurveySchema):
...
             raise ValueError('Engagement already published')
-        return SurveyModel.update_survey(data)
+
+        updated_survey = SurveyModel.update_survey(data)
+        ReportSettingService.refresh_report_setting({
+            'id': updated_survey.id,
+            'form_json': updated_survey.form_json,
+        })
+
+        return updated_survey
```
saravanpa-aot, line 163: Plz check this call with @VineetBala-AOT 
i am not sure if this is needed...his design was to refresh only when the user visits the page

jadmsaadaot: Thank you for bringing this up. I checked with @VineetBala-AOT. He mentioned that his design is on the survey build page. When we save the survey we do one api call to save the survey and another after to save the settings. I discussed with him to not have another api call there and move the saving the settings to the survey service in the create survey method. We will always create report settings from survey that's why I called it here and eliminates handling error happening on one api call and not the other. Vineet was positive about this.

@VineetBala-AOT thoughts?

met-api/tests/utilities/factory_utils.py
```diff
@@ -301,3 +302,17 @@ def factory_engagement_slug_model(eng_slug_info: dict = TestEngagementSlugInfo.s
     )
     slug.save()
     return slug
+
+
+def factory_survey_report_setting_model(eng_slug_info: dict = TestReportSettingInfo.report_setting_1):
```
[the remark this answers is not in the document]
jadmsaadaot: it is not, it was an oversight from me. Changed to relevant name. Thanks for noticing

met-api/src/met_api/services/survey_service.py
saravanpa-aot: hm..challenge is it might create entries for every survey..
the survey settings might not be used much..only for a few engagements might utilise this feature..like i think may be less than 10% of the survey we will have this setting customised

saravanpa-aot (approved):

PR #1828, opened 2023-07-11
Add video widget admin side component and resource

jadmsaadaot:
*Description of changes:*

 - Add video widget to option cards
- Add video widget drawer form
- Add create, get and update video widget resources


met-api/migrations/versions/47fc88fe0477_video_widget.py
```diff
@@ -0,0 +1,52 @@
...
+    sa.ForeignKeyConstraint(['widget_id'], ['widget.id'], ondelete='CASCADE'),
+    sa.PrimaryKeyConstraint('id')
+    )
+    widget_type_table = sa.table('widget_type',
+        sa.Column('id', sa.Integer),
+        sa.Column('name', sa.String),
+        sa.Column('description', sa.String))
+
+    op.bulk_insert(widget_type_table, [
+        {'id': 7, 'name': 'Video', 'description': 'Add a link to a hosted video and link preview'}
```
saravanpa-aot, line 40: in one of the alembic scripts , i haev a better of handling the increment numbers.. can you look up that?
basically it reads value from the sequnce and does a plus one

met-web/src/services/widgetService/VideoService/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+    try {
+        const url = replaceAllInURL({
+            URL: Endpoints.VideoWidgets.UPDATE,
+            params: {
+                widget_id: String(widget_id),
+                video_widget_id: String(video_widget_id),
+            },
+        });
+        const response = await http.PatchRequest<VideoWidget>(url, data);
+        if (response.data) {
```
saravanpa-aot, line 55: return response.data || Promise.reject(' etc

works? the if statement looks like long

met-web/src/services/widgetService/VideoService/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export const postVideo = async (widget_id: number, data: PostVideoRequest): Promise<VideoWidget> => {
+    try {
+        const url = replaceUrl(Endpoints.VideoWidgets.CREATE, 'widget_id', String(widget_id));
+        const response = await http.PostRequest<VideoWidget>(url, data);
+        if (response.data) {
+            return response.data;
+        }
+        return Promise.reject('Failed to create video widget');
```
saravanpa-aot, line 30: same like below..may be use a || to avoid if 
make sure it works please

saravanpa-aot (approved):

met-api/migrations/versions/47fc88fe0477_video_widget.py
jadmsaadaot: That's definitely a smarter way to do it. I just choose to hard code 7 here because it is hard coded in the front end:
image

And types are hard coded in the backend in the constants in WidgetType which is used for the tests. I wasn't comfortable leaving it up for a calculation although it will reach the same result of 7. What do you think

image

PR #1871, opened 2023-07-19
Button styling update

djnunez-aot:
-Remove form io button radius:0 as it prevents us from styling their components


jadmsaadaot:
"-Remove form io button radius:0 as it prevents us from styling their components"

their? which components?

djnunez-aot:
@jadmsaadaot The ones that show up when we display a survey. In this case it's the next and previous button. This pr is made because there's some small hover artifacts since form io scss overrides some styles for the buttons.

PR #1881, opened 2023-07-24
Adding marker to map chart

VineetBala-AOT:
*Description of changes:*
- Added a circular marker to the map chart while exporting.

image


jadmsaadaot (approved):
No chance it could be like the pin marker?

Issue #1899, opened 2023-07-27
Reviewer can access engagements & surveys through hacking the URL


marieanik:
All the followings could hack the URL to access engagements/survey they shouldn't have access to:

Geoff as Reviewer could view all the surveys attached to engagements he wasn't assigned to.
MA as Viewer could access engagements and surveys in draft and hidden surveys.
Taya as TM could access draft engagements she isn't assigned to and hidden surveys.

*Should be fixed by doing a pass on endpoints and pages and add role restrictions

TayaLee8:
Passing in TEST: can still hack into closed/open engagements (desired behaviour), but cannot hack hidden/draft engagements

PR #1968, opened 2023-08-04
auth check added

saravanpa-aot:
https://github.com/bcgov/met-public/issues/1899

*Description of changes:*

1. Added auth_check for survey GET and engagement GET
2. Added tests for the same

### DIFF
--- a/met-api/src/met_api/services/engagement_service.py
+++ b/met-api/src/met_api/services/engagement_service.py
@@ -31,15 +31,23 @@ class EngagementService:
     otherdateformat = '%Y-%m-%d'
 
     @staticmethod
-    def get_engagement(engagement_id, user_id) -> EngagementSchema:
+    def get_engagement(engagement_id) -> EngagementSchema:
         """Get Engagement by the id."""
         engagement_model: EngagementModel = EngagementModel.find_by_id(engagement_id)
 
         if engagement_model:
-            if user_id is None \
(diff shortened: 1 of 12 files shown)


met-api/src/met_api/services/engagement_service.py
```diff
@@ -31,15 +31,23 @@ class EngagementService:
...
     @staticmethod
-    def get_engagement(engagement_id, user_id) -> EngagementSchema:
+    def get_engagement(engagement_id) -> EngagementSchema:
         """Get Engagement by the id."""
         engagement_model: EngagementModel = EngagementModel.find_by_id(engagement_id)
 
         if engagement_model:
-            if user_id is None \
+            if TokenInfo.get_id() is None \
                     and engagement_model.status_id not in (Status.Published.value, Status.Closed.value):
```
jadmsaadaot, line 40: Do you think (Status.Published.value, Status.Closed.value) should be passed as argument. This condition might change I think. (Status.Published.value, Status.Closed.value)  could be the default maybe

saravanpa-aot: i think i can make it to a separate method and improve how it looks like i think..
If the calling method has to pass the status , it might be like the calling method decides the security..so someone might call the method not knowing whats the status and override... thats the challenege

jadmsaadaot (approved):

PR #1994, opened 2023-08-10
Implement membership revoke and versioning

jadmsaadaot:
*Description of changes:*
- Add resource endpoint and service methods to revoke and reinstate
- Create a new version of membership when revoking and reinstating


jadmsaadaot:
I will follow up this PR with unit tests in back and front

met-api/src/met_api/models/membership.py
```diff
@@ -24,45 +24,80 @@ class Membership(BaseModel):
...
+            engagement_id=engagement_id,
+            user_id=user_id,
+            status=new_membership.get('status'),
+            type=new_membership.get('type'),
+            revoked_date=new_membership.get('revoked_date', None),
+            is_latest=True,
+            version=latest_membership.version + 1
+        )
+        new_membership.save()
+        return new_membership
```
saravanpa-aot, line 103: this is very clean..
the only improvement i can think is of using an update query directly and return a scalar.. that will avoid a full table fetch
https://stackoverflow.com/questions/45842715/how-to-use-returning-for-query-update-with-sqlalchemy

something like update(). returning(version)

But this code is clean, i almost dont want to change it at all.. , so i will the decision to you

met-web/src/services/membershipService/index.tsx
```diff
@@ -38,3 +38,30 @@ export const getMembershipsByUser = async ({
...
+    membership_id: number,
+): Promise<EngagementTeamMember> => {
+    const url = replaceAllInURL({
+        URL: Endpoints.EngagementTeamMembers.REINSTATE,
+        params: {
+            engagement_id: String(engagement_id),
+            membership_id: String(membership_id),
+        },
+    });
+    const responseData = await http.PutRequest<EngagementTeamMember>(url);
```
saravanpa-aot, line 65: on another note , when we think about the revoking , do we generally revoke a membership from an engagement or user from engagement 

should the rest point accept , user_id and engaggement_id ? so thats its the whole users membership getting revoked instead of membership_id and engagemnet_id? I am not too sure about whats right.

saravanpa-aot (commented):
very nearly done..a few comments added

met-web/src/services/membershipService/index.tsx
jadmsaadaot: I am thinking we are revoking a membership, I thought about this. Either membership.id or engagement.id. The reason why I chose engagement.id is to for sure get the is_latest of the assignment for this user on this engagement and make it is_latest false. so I can add another membership with this engagement and user id. In my head user_id and engagement_id are like custom primary key

jadmsaadaot: yes if I fetch membership.id I get that membership. but with user_id and engagement_id I know I get the latest membership for that combined key

jadmsaadaot: Yes changed to user_id and engagement_id, I think we should treat that like custom combined primary key. I was thinking between the two and wasn't sure as well.

met-api/src/met_api/models/membership.py
jadmsaadaot: I can change it to that, it is not a very frequent call. I would value clarity and easiness to change for the maintainer here

met-api/src/met_api/services/membership_service.py
```diff
@@ -124,3 +125,49 @@ def get_assigned_engagements(user_id):
...
+            raise ValueError('Membership already active.')
+
+        new_membership_details = {
+            'engagement_id': engagement_id,
+            'user_id': membership.user_id,
+            'status': MembershipStatus.ACTIVE.value,
+            'type': membership.type,
+        }
+        new_membership = MembershipModel.create_new_version(engagement_id, membership.user_id, new_membership_details)
+        return new_membership
```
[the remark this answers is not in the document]
jadmsaadaot: updated

met-api/src/met_api/resources/engagement_members.py
```diff
@@ -76,3 +76,37 @@ def get(engagement_id, user_id):  # pylint: disable=unused-argument
...
+    @staticmethod
+    @cross_origin(origins=allowedorigins())
+    @_jwt.has_one_of_roles([Role.EDIT_MEMBERS.value])
+    def put(engagement_id, membership_id):
+        """Reinstate engagement membership for a user."""
+        try:
+            membership = MembershipService.reinstate_membership(engagement_id, membership_id)
+            return MembershipSchema().dump(membership), HTTPStatus.OK
+        except BusinessException as err:
+            return {'message': err.error}, err.status_code
```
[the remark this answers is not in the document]
jadmsaadaot: Took this approach, thanks for the suggestion

met-api/src/met_api/models/membership.py
saravanpa-aot: yea..i almost dont wanna touch this code,.

saravanpa-aot (approved):

saravanpa-aot (approved):

PR #2005, opened 2023-08-14
Add listing tables to redux to save table state

djnunez-aot:
-Add redux state for each listing table


djnunez-aot (commented):
Approach Changed to Query Params so state is only saved when user goes back on the browser

Issue #1761, opened 2023-06-21
Configure Bucket for Tenants


saravanpa-aot:

VineetBala-AOT:
Verified the new logic is working in dev.

PR #2011, opened 2023-08-14
Add dynamic s3 config

jadmsaadaot:
https://github.com/bcgov/met-public/issues/1761

*Description of changes:*
- Add a util to dynamically get s3 config depending on tenant

### DIFF
--- a/met-api/src/met_api/config.py
+++ b/met-api/src/met_api/config.py
@@ -51,7 +51,6 @@ def get_s3_config(key: str):
 
     :raise: KeyError: if an unknown configuration is requested
     """
-    print(">>>>>")
     tenant_short_name = g.get('tenant_name', None)
     if not tenant_short_name:
         return _Config.S3_CONFIG['DEFAULT'][key]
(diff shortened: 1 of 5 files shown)


met-api/src/met_api/config.py
```diff
@@ -82,6 +106,17 @@ class _Config():  # pylint: disable=too-few-public-methods
...
+        'DEFAULT': {
+            'S3_BUCKET': os.getenv('S3_BUCKET'),
+            'S3_ACCESS_KEY_ID': os.getenv('S3_ACCESS_KEY_ID'),
+            'S3_SECRET_ACCESS_KEY': os.getenv('S3_SECRET_ACCESS_KEY'),
+            'S3_HOST': os.getenv('S3_HOST'),
+            'S3_REGION': os.getenv('S3_REGION'),
+            'S3_SERVICE': os.getenv('S3_SERVICE')
+        }
+    }
+
```
saravanpa-aot, line 119: can we please add a section of GDX or some tenant so that there is an example in the config?

saravanpa-aot (approved):

met-api/src/met_api/config.py
jadmsaadaot: We had a discussion, where you confirmed we will not do this

PR #2126, opened 2023-09-05
Save engagement notification message

djnunez-aot:
-dont display 'Engagement Metadata created successfully' when preview button or save button is pressed on engagement form


jadmsaadaot (changes requested):
Code smells

jadmsaadaot (changes requested):
Do we wanna keep updating metadata when we update an engagement?

We can do something similar to what we are doing for settings. At the service level when we create an engagement we create the metadata. And updating it can happen separately on the settings tab after create
image

djnunez-aot:
@jadmsaadaot that's a great idea actually ill see if i can move the update metadata part to the backend

jadmsaadaot (approved):

Issue #1327, opened 2023-03-07
form.io edit form pop-up becomes stuck if background page is changed


TayaLee8:

TayaLee8:
Talked to Jad about this and confirmed it was a bug and needed a ticket @jadmsaadaot

TayaLee8:
**On MacBook**
1. User goes to add a form component to a survey
2. After dropping the form component, a pop-up appears
3. User accidently navigates back to the previous **background** screen (two finger swipe on MacBook)
4. The pop-up becomes stuck, the user cannot save, cancel, or remove their form component and is forced to close the tab 

https://user-images.githubusercontent.com/122818332/226674157-8844c7eb-300c-47a3-8984-554d9c3305f8.mov

carlygreenehill:
@jadmsaadaot + @TayaLee8  + @marieanik  --- did we fix this bug or does it still need our attention? Please reply to this comment if it is still a thing. 

I'm ranking this a $$$ (big money, high priority) bug if it's still active because it might lead to Survey Creators losing all their work.

marieanik:
gng

PR #2161, opened 2023-09-11
Formio modal bug fix

djnunez-aot:
Issue #: https://github.com/bcgov/met-public/issues/1327
-Add a listener that unrenders the formio modal on switching routes

### DIFF
--- a/met-web/src/components/FormioModalListener/index.tsx
+++ b/met-web/src/components/FormioModalListener/index.tsx
@@ -1,0 +1,17 @@
+import { useEffect } from 'react';
+import { useLocation } from 'react-router-dom';
+
+const FormioModalListener = () => {
+    const { pathname } = useLocation();
+
+    useEffect(() => {
+        const formioModal = document.querySelector('.formio-dialog');
+        if (formioModal) {
+            formioModal.remove();
+        }
+    }, [pathname]);
+
+    return null;
+};
+
+export default FormioModalListener;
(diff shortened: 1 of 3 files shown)


met-web/src/components/scrollToTop/index.tsx
```diff
@@ -6,6 +6,10 @@ const ScrollToTop = () => {
 
     useEffect(() => {
```
jadmsaadaot, line 7: I like this, can we please have it in its own separate component though. We can name it something relevant and specific to Formio. FormioModalListener... maybe

jadmsaadaot (changes requested):
Do change in separate Formio specific component.

met-web/src/components/scrollToTop/index.tsx
djnunez-aot: For sure ill do that now

met-web/src/components/FormioModalListener/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+    const { pathname } = useLocation();
+
+    useEffect(() => {
+        const formioModal = document.querySelector('.formio-dialog');
+        if (formioModal) {
+            formioModal.remove();
+        }
+    }, [pathname]);
+
+    return null;
```
jadmsaadaot, line 14: why the return null? Is it needed?

PR #2179, opened 2023-09-13
Fix for New Staff users were getting blank status

saravanpa-aot:
Issue #: https://github.com/bcgov/met-public/issues/

*Description of changes:*

Server_default was not working;probably bcoz there is no database constraint..easiser option is to work with default.so switching back to it

### DIFF
--- a/met-api/src/met_api/models/staff_user.py
+++ b/met-api/src/met_api/models/staff_user.py
@@ -33,7 +33,7 @@ class StaffUser(BaseModel):
     email_address = Column(db.String(100), nullable=True)
     contact_number = Column(db.String(50), nullable=True)
     external_id = Column(db.String(50), nullable=False, unique=True)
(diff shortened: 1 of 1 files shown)


djnunez-aot (approved):

jadmsaadaot (approved):

met-api/src/met_api/models/staff_user.py
```diff
@@ -33,7 +33,7 @@ class StaffUser(BaseModel):
     email_address = Column(db.String(100), nullable=True)
     contact_number = Column(db.String(50), nullable=True)
     external_id = Column(db.String(50), nullable=False, unique=True)
-    status_id = db.Column(db.Integer, ForeignKey('user_status.id'), nullable=False, server_default='1')
+    status_id = db.Column(db.Integer, ForeignKey('user_status.id'), nullable=False, default=1)
```
jadmsaadaot, line 36: What about the migration that was created?

Issue #1088, opened 2023-01-11
Export Issue Tracking Table - Upgrades - Superuser version


saravanpa-aot:
The name of the file generated will be in the format: INTERNAL ONLY - Engagement Name yyyy-mm-dd hh:mm 
Current columns and action required:

Comment No. - Keep
Submitted - Keep
Author - Remove
Location - Remove
Comment - Replace with a column for each Single-line/multiple-line question/comment. Column headers will be the label. Include all questions, even they are not showed on the public report)
Attachment - Remove
**Add:** Status (status of the comment in MET: Approved, Rejected, Needs further Review
Published - Change to published Date (the date the comment was approved/published)
Deferral Notes - Remove
Rejection Note - Change to "Reason for Rejection"
Reviewer - keep
Project - Keep (take from Metadata)
PCP Title - Remove (it's in the name of the file)
Export Date - Remove (it's in the name of the file)

Example of what the tracking table  will look like:
[GreenHarbor Resort Expansion - 2023-08-22 20_25_26.xlsx](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaHRVIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--26368ac3d670acaad58bb3afdb45bdad1d1faa29/GreenHarbor%20Resort%20Expansion%20-%202023-08-22%2020_25_26.xlsx)

carlygreenehill:
Here's what my notes say:

Column C - Author - DELETE
Column D - Location - DELETE
Column E - Comment - KEEP
Column F - Attachment - DELETE

Column G - Published  = date published
Column H - Deferral Note = DELETE (I think this could the space for Internal Notes, but don't see it's purpose in the export)
Column I - Rejection Note = might be useful for tracking what % of comments we have to reject and the reasons why, not sure if it should be on this .csv tho or if it can be handled in Web Analytics

Column K - Project - make sure pulled from metadata 
Column L - PCP Title - pull from metadata? Or from Engagement Name? 
Column M - Export Date - I didn't see a problem keeping it, but now that I read your comments, you're right, this isn't really needed


Can we add a column that indicates if a comment was revised?
Ask about if we do/can keep data about resubmissions?
What is the difference between COMMENT # and Comment ID? (.CSV vs. in- app comment review)
Can we add a column or some way of indicating what the Question was (aka the label for the multi line)

[...]

PR #2192, opened 2023-09-14
Upgrades to Issue Tracking Table

VineetBala-AOT:
Issue #: https://github.com/bcgov/met-public/issues/1088

*Description of changes:*
- Updated the template for Issue Tracking Table
- Added logic to have dynamic title based on the number of comment type questions


met-api/src/met_api/services/comment_service.py
```diff
@@ -158,32 +161,90 @@ def extract_comments_from_survey(cls, survey_submission: SubmissionSchema, surve
...
+            engagement = SurveyModel.find_by_id(survey_id)
+            one_of_roles = (
+                MembershipType.TEAM_MEMBER.name,
+                Role.EXPORT_TO_CSV.value
+            )
+            authorization.check_auth(one_of_roles=one_of_roles, engagement_id=engagement.engagement_id)
+            comments = Comment.get_comments_by_survey_id(survey_id)
+            metadata_model = EngagementMetadataModel.find_by_id(engagement.engagement_id)
+            project_name = metadata_model.project_metadata.get('project_name') if metadata_model else None
+        except BusinessException as err:
```
jadmsaadaot, line 174: should this try catch be here? Wouldn't the one in the resource be able to handle it

jadmsaadaot (changes requested):
Should the try catch be at the resource level instead

met-api/src/met_api/services/comment_service.py
VineetBala-AOT: yes, removed it.

jadmsaadaot (approved):

Issue #2160, opened 2023-09-11
Change Page title (text appearing on browser tab) to EPIC.engage for the EAO instance


marieanik:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBbzZKIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--9ec7556b51d0c490725471c342b086815a152bcf/Screenshot%202023-09-11%20at%2012.03.02%20PM.png)

PR #2201, opened 2023-09-15
Add tenant name to page name

djnunez-aot:
Issue #2160 

-add tenant name to page name on loading tenant

### DIFF
--- a/met-web/src/App.tsx
+++ b/met-web/src/App.tsx
@@ -94,10 +94,16 @@ const App = () => {
 
     const getTranslationFile = async () => {
         try {
-            const translationFile = await import(`./locales/${language}/${tenant.id}.json`);
+            const translationFile = await import(`./locales/${language}/${basename}.json`);
+            if (translationFile && translationFile.header && translationFile.header.title) {
+                document.title = translationFile.header.title;
+            }
             return translationFile;
         } catch (error) {
             const defaultTranslationFile = await import(`./locales/${language}/default.json`);
+            if (defaultTranslationFile && defaultTranslationFile.header && defaultTranslationFile.header.title) {
+                document.title = defaultTranslationFile.header.title;
+            }
             return defaultTranslationFile;
         }
     };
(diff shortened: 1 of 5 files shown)


saravanpa-aot (dismissed):
There is a localisation file which should be used for this..

met-web/src/App.tsx
```diff
@@ -94,10 +94,12 @@ const App = () => {
 
     const getTranslationFile = async () => {
         try {
-            const translationFile = await import(`./locales/${language}/${tenant.id}.json`);
+            const translationFile = await import(`./locales/${language}/${basename}.json`);
+            document.title = translationFile?.header?.title || document.title;
```
jadmsaadaot, line 98: ok so maybe we need to leverage React Helmet to do something like this, it's probably cleaner. Can you check the library

met-web/src/App.tsx
```diff
@@ -94,10 +94,12 @@ const App = () => {
 
     const getTranslationFile = async () => {
         try {
-            const translationFile = await import(`./locales/${language}/${tenant.id}.json`);
```
jadmsaadaot, line 97: also why the change from tenant.id to basename, this overrides my changes in another PR for the single tenancy/ multi tenancy task

met-web/src/App.tsx
```diff
@@ -95,9 +97,13 @@ const App = () => {
     const getTranslationFile = async () => {
         try {
             const translationFile = await import(`./locales/${language}/${tenant.id}.json`);
+            const title = translationFile?.header?.title || document.title;
```
jadmsaadaot, line 100: that's not how we get values from the translationFile. Check     const { t: translate } = useAppTranslation(); in the landingComponent for example

jadmsaadaot (changes requested):
use const { t: translate } = useAppTranslation();

jadmsaadaot (approved):

PR #2206, opened 2023-09-15
check if obj on hasKey is object

jadmsaadaot:
*Description of changes:*
- Make hasKey utility function safe when object is falsey


saravanpa-aot (approved):
the return type looks like boolean instead of

PR #2302, opened 2023-10-06
Fix selected question should be highlighted

jadmsaadaot:
Issue #:

*Description of changes:*
- Use Index value to highlight the selected question


saravanpa-aot (approved):

jadmsaadaot:
The code smell can be ignored, it is a problem using index as key if the list might be reordered by design. But we will not reorder it 

        <li key={index}>

Issue #408, opened 2022-08-09
Date published doesn't always show up on Engagement listing


marieanik:
I created and published the last two engagements. Both are published but the date for the first one doesn't show up.
Logged out, came back in MET and now the date doesn't show up for the second one as well.

![Screen Shot 2022-08-09 at 2.26.40 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/110d4839-9389-4aaa-add3-edb043cc12a8)
![Screen Shot 2022-08-09 at 3.31.22 PM.png](https://images.zenhubusercontent.com/6261bc1a04c6a94c2edadddd/41a9d25c-cf78-47c6-ab8b-4b10b15bc370)

saravanpa-aot:
@marieanik can you verify this?

PR #2333, opened 2023-11-06
Feature: SSO Migration - DESENG #408

NatSquared:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-408

*Description of changes:*

> **Feature**: .env var audit and cleanup - [🎟️DESENG-414](https://apps.itsm.gov.bc.ca/jira/browse/DESENG-414) 
> - Update sample.env files to properly reflect the current state of the project and document where to download credentials securely
> *This ticket is not closed by this version*


met-api/src/met_api/__init__.py
```diff
@@ -128,8 +128,15 @@ def build_cache(app):
...
 
-    def get_roles(a_dict):
-        return a_dict['realm_access']['roles']  # pragma: no cover
-
+    def get_roles(token_info):
+        """
+        Consumes a token_info dictionary and returns a list of roles.
+        Uses a configurable path to the roles in the token_info dictionary.
+        """
+        role_access_path = app_context.config['JWT_ROLE_CLAIM']
```
Baelx, line 136: Nice touch with the customizable path. Will be nice to have while we're in this transition period between auth providers.

met-api/src/met_api/models/base_model.py
```diff
@@ -67,16 +67,10 @@ def flush(self):
...
-        """Save and flush."""
-        return self.flush()
-
     def save(self):
         """Save and commit."""
-        self._set_tenant_id()
-        db.session.add(self)
-        db.session.flush()
-        db.session.commit()
+        self.flush()
```
Baelx, line 72: Does `self` access `db.session` in this instance?

NatSquared: The first argument to a method of a class (called [`self`](https://www.w3schools.com/python/gloss_python_self.asp) by convention) always receives a reference to the class itself (in this case, `BaseModel`). So `self.flush()` is the same as [`BaseModel.flush()`](https://github.com/bcgov/met-public/blob/418208247e4ace26a8006cb3175db59ef048abc9/met-api/src/met_api/models/base_model.py#L63), which behind the scenes gets called as something like BaseModel.flush(self=BaseModel)

met-web/sample.env
```diff
@@ -1,9 +1,10 @@
-# Keycloak auth endpoint
-REACT_APP_KEYCLOAK_URL=https://dev.loginproxy.gov.bc.ca/auth
-REACT_APP_KEYCLOAK_REALM=standard
 
-# Resource identifier for the Keycloak client
-REACT_APP_KEYCLOAK_CLIENT=modern-engagement-tools-4787
+# Keycloak auth
+# Copy from 'GDX MET web (public)-installation-*.json'
```
Baelx, line 3: Nice helpful notes to aid in gathering those env variables

met-api/src/met_api/services/authorization.py
```diff
@@ -21,12 +21,13 @@ def check_auth(**kwargs):
...
     skip_tenant_check = current_app.config.get('IS_SINGLE_TENANT_ENVIRONMENT')
     user_from_context: UserContext = kwargs['user_context']
+    user_from_db = StaffUserModel.get_user_by_external_id(user_from_context.sub)
     token_roles = set(user_from_context.roles)
     permitted_roles = set(kwargs.get('one_of_roles', []))
     has_valid_roles = token_roles & permitted_roles
     if has_valid_roles:
         if not skip_tenant_check:
-            user_tenant_id = user_from_context.tenant_id
+            user_tenant_id = user_from_db.tenant_id
```
Baelx, line 30: If I understand correctly, you switched accessing the "current" user from app context to pulling that user from the DB. Was the former solution causing issues?

Baelx:
APPROVED: Nice work on this! I especially appreciate the comments and detailed changelog and PR notes. Looks good :)

PR #2337, opened 2023-11-21
bugfix/deseng429: Removed outdated service class.

jareth-whitney:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-429

- Removed outdated service class (engagement_slug.py, replaced by engagement_slug_class.py)


CHANGELOG.MD
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](https://semver.org/).
 
+## v1.1.1 - 2023-11-21
```
Baelx, line 5: This would be the correct semver level, but in my opinion, we should only increment the version number when we do a release(i.e. push to production). This has been what we've done in the past. 

My understanding is that we haven't deployed any of our work yet. Is that correct? What are your thoughts on incrementing outside of a deployment? The alternative would be to keep grouping all changes under a single heading until that version is deployed.

jareth-whitney: Yes I was wondering about this as well. Our version control won't be as strict without the version tagging system that we use on WP deployments. I have merged with gdx-main before, I believe that was done after the bc-sans font update. So if you think we should downgrade to version 1.1.0 and leave anything under that, I'd be ok with that.

Baelx: I was thinking the same thing regarding downgrading to 1.1.0. sounds good!

NatSquared (approved):
Looks good! I thought we might end up being able to just get rid of that .py file. Also, I agree with @Baelx's suggestion re: version numbering
Approved 🎊

PR #2340, opened 2023-12-05
Feature/deseng441 - Remove unused engagement metadata

Baelx:
I've gone about removing the current metadata fields in a few ways:
__Frontend__:
- Outright remove the fields from the UI
- Remove and modify types that include references to those fields
- Remove or comment-out references to engagement "metadata." Since we'll be using metadata again, I tried to leave references to it as intact as possible for now.

__Backend__:
- Remove references to the old metadata fields
- Comment-out metadata classes/function calls in case they're useful when adding back metadata to the app. I've left TODO comments at each location


NatSquared (approved):

met-api/src/met_api/models/engagement.py
```diff
@@ -263,29 +264,7 @@ def _filter_by_internal(query, search_options):
...
-            query = query.filter(EngagementMetadataModel.project_metadata['application_number']
-                                 .astext.ilike(f'%{application_number}%')) \
-                .params(val=application_number)
-
-        if client_name := search_options.get('client_name'):
-            query = query.filter(EngagementMetadataModel.project_metadata['client_name']
-                                 .astext.ilike(f'%{client_name}%')) \
-                .params(val=client_name)
-
+        # TODO: Populate or remove this method dependent on changes resulting from adding the new Engagement metadata
```
jareth-whitney, line 267: Thanks for leaving these TODOs for the next phase of this metadata project.

met-api/src/met_api/resources/engagement.py
```diff
@@ -90,11 +90,6 @@ def get():
                 'created_to_date': args.get('created_to_date', None, type=str),
                 'published_from_date': args.get('published_from_date', None, type=str),
                 'published_to_date': args.get('published_to_date', None, type=str),
-                'project_type': args.get('project_type', None, type=str),
-                'project_id': args.get('project_id', None, type=str),
```
jareth-whitney, line 94: Is it ok to remove the project_id? Is it a primary key?

met-api/src/met_api/services/project_service.py
```diff
@@ -28,11 +28,7 @@ def update_project_info(eng_id: str) -> EngagementModel:
             engagement_metadata: EngagementMetadataModel
             engagement, engagement_metadata = ProjectService._get_engagement_and_metadata(eng_id)
 
-            if not (project_id := engagement_metadata.project_id):
-                # EPIC is not interested in the data without project Id.So Skip the process.
```
jareth-whitney, line 32: Will the process be skipped every time because project_id is never available? If so, should more code be removed?

met-web/src/components/engagement/form/EngagementFormTabs/Settings/EngagementInformation.tsx
```diff
@@ -19,90 +17,13 @@ const EngagementInformation = () => {
...
-                <MetLabel>Application #</MetLabel>
-                <TextField
-                    id="application-number"
-                    name="application_number"
-                    value={project_metadata.application_number}
-                    variant="outlined"
-                    fullWidth
-                    onChange={handleChangeMetadata}
-                />
+                <MetHeader4 bold>Engagement Metadata</MetHeader4>
```
jareth-whitney, line 26: Are we keeping this header because we will have different metadata later?

met-web/src/config.ts
```diff
@@ -64,20 +61,6 @@ const KC_CLIENT = getEnv('REACT_APP_KEYCLOAK_CLIENT');
 const KC_REALM = getEnv('REACT_APP_KEYCLOAK_REALM');
 const KC_ADMIN_ROLE = getEnv('REACT_APP_KEYCLOAK_ADMIN_ROLE');
 
-// App constants
-const ENGAGEMENT_PROJECT_TYPES: string[] = getEnv(
```
jareth-whitney, line 68: lol yikes

jareth-whitney (approved):
Looks like the challenge here was knowing what to strip without breaking anything. Thanks for adding the TODOs and I've added a few comments just in case things need to be addressed. Approving PR.

met-api/src/met_api/resources/engagement.py
Baelx: Good question! From reading the user documentation and inspecting the code, it looks like there isn't a table for projects but that the user just enters the ID manually. So no, I don't think they're a primary key.

My guess is that "projects" are domain-specific to EAO and they're managed outside of the MET app. The ticket wasn't explicit about the user of "projects" and their IDs in the Engagement version of MET so I left some instances of project_id intact until we get direction from the PO on that.

met-api/src/met_api/services/project_service.py
Baelx: I think most likely yes and yes to your question. The ticket didn't give explicit direction for how to handle "Projects" so I decided to leave this for now until we get a "Remove project functionality from MET" ticket

met-web/src/components/engagement/form/EngagementFormTabs/Settings/EngagementInformation.tsx
Baelx: That was my thought, yes! And the ticket implies new metadata will go here too

met-web/src/config.ts
Baelx: Yikes at the hard-coded nature of project types? :P

jareth-whitney: Yes exactly

PR #2358, opened 2024-01-12
DESENG-436: MET - Rewrite unit tests

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-436

*Description of changes:*
- Resolved API unit test failures on local runs by updating the test suite to pre-create a user before execution.


met-api/sample.env
```diff
@@ -124,6 +124,8 @@ DATABASE_TEST_NAME=
 DATABASE_TEST_HOST=
 DATABASE_TEST_PORT=
 
+KEYCLOAK_TEST_BASE_URL="http://localhost:8081"
```
Baelx, line 127: So does this mean one needs to run an instance of keycloak locally just to be able to test?

met-api/src/met_api/services/widget_timeline_service.py
```diff
@@ -28,23 +33,48 @@ def create_timeline(widget_id: int, timeline_details: dict):
...
 
-        return WidgetTimelineModel.update_timeline(timeline_id, timeline_data)
+    @staticmethod
+    def _check_update_timeline_auth(first_event):
+        eng_id = first_event.get('engagement_id')
+        authorization.check_auth(one_of_roles=(MembershipType.TEAM_MEMBER.name,
+                                               Role.EDIT_ENGAGEMENT.value), engagement_id=eng_id)
+
+    @staticmethod
+    def _update_widget_timeline(widget_timeline: WidgetTimelineModel, timeline_data: dict):
```
Baelx, line 71: Why was the update method for timelines moved to the service from the model? 

I've noticed a pattern in this app where models sometimes contain all CRUD methods and sometimes contain only some of them, with the service taking over. Do you happen to know the reason for this? In my mind, we'd strive to have all method handlers in the same location (perhaps the model). If there's a good reason for this that I'm missing, I'd be curious to hear it!

met-api/src/met_api/services/widget_timeline_service.py
```diff
@@ -56,13 +86,13 @@ def _create_timeline_model(widget_id: int, timeline_data: dict):
...
             timeline_model.events.append(
                 TimelineEventModel(
-                    widget_id = widget_id,
-                    engagement_id = event.get('engagement_id'),
-                    timeline_id = event.get('timeline_id'),
-                    description = event.get('description'),
-                    time = event.get('time'),
-                    position = event.get('position'),
-                    status = event.get('status'),
+                    widget_id=widget_id,
```
Baelx, line 89: Did the linter ask for the removal of these spaces, is there a rule associated with this relevant to classes? I personally prefer when they're spaced out - to me that's more readable.

met-api/tests/conftest.py
```diff
@@ -169,3 +172,51 @@ def docker_compose_files(pytestconfig):
...
+    """Set up a user with the reviewer role."""
+    staff_info = dict(TestUserInfo.user_staff_1)
+    user = factory_staff_user_model(user_info=staff_info)
+    claims = copy.deepcopy(TestJwtClaims.reviewer_role.value)
+    claims['sub'] = str(user.external_id)
+
+    return user, claims
+
+
+# Fixture for setting up user and claims for an team member
```
Baelx, line 201:
```suggestion
# Fixture for setting up user and claims for a team member
```

met-api/tests/unit/api/test_engagement.py
```diff
@@ -95,19 +101,23 @@ def test_tenant_id_in_create_engagements(client, jwt, session):  # pylint:disabl
     assert response.json['tenant_id'] == str(tenant_2.id)
 
 
-@pytest.mark.parametrize('role', [TestJwtClaims.no_role, TestJwtClaims.public_user_role])
-def test_add_engagements_invalid(client, jwt, session, role):  # pylint:disable=unused-argument
+@pytest.mark.parametrize('engagement_info', [TestEngagementInfo.engagement1])
+def test_add_engagements_invalid(client, jwt, session, engagement_info,
+                                 setup_unprivileged_user_and_claims):  # pylint:disable=unused-argument
     """Assert that an engagement can not be POSTed without authorisaiton."""
```
Baelx, line 107:
```suggestion
    """Assert that an engagement can not be POSTed without authorization."""
```

met-api/tests/unit/api/test_survey.py
```diff
@@ -39,9 +39,11 @@
 
 
 @pytest.mark.parametrize('survey_info', [TestSurveyInfo.survey1])
-def test_create_survey(client, jwt, session, survey_info):  # pylint:disable=unused-argument
+def test_create_survey(client, jwt, session, survey_info,
+                       setup_admin_user_and_claims):  # pylint:disable=unused-argument
```
Baelx, line 43: Do you know if we need to keep the unused argument around instead of adding a rule to ignore it?

met-web/tests/unit/components/engagement/form/edit/EngagementForm.Edit.One.test.tsx
```diff
@@ -203,12 +203,13 @@ describe('Engagement form page tests', () => {
...
 
-        expect(screen.getByText('Engagement Information')).toBeInTheDocument();
         expect(screen.getByText('Internal Engagement')).toBeInTheDocument();
         expect(screen.getByText('Send Report')).toBeInTheDocument();
+        expect(screen.getByText('Link to Public Engagement Page')).toBeInTheDocument();
+        expect(screen.getByText('Link to Public Dashboard Report')).toBeInTheDocument();
     });
 
-    test('Can move to links tab', async () => {
+    test('Can move to additional details tab', async () => {
```
Baelx, line 212: 👍

met-web/tests/unit/components/engagement/form/edit/EngagementForm.Edit.Two.test.tsx
```diff
@@ -145,7 +145,7 @@ describe('Engagement form page tests', () => {
 
         await waitFor(() => {
             expect(screen.getByDisplayValue('Test Engagement')).toBeInTheDocument();
-            expect(container.querySelector('span.MuiSkeleton-root')).toBeNull();
+            expect(container.querySelector('span.MuiSkeleton-root'));
```
Baelx, line 148: Just curious: what does calling `expect` without a matcher function checks for?

Baelx (approved):
Nice work! There's not much to quibble with but I am curious about a few things. If you could please answer my questions before proceeding, I'd appreciate it

met-api/sample.env
VineetBala-AOT: No need to run a separate instance of Keycloak. The test utilizes a Docker Compose file, and the Keycloak instance is automatically set up when executing the pytest command locally.

Baelx: Oh okay so if I'm understanding: a limited, temporary instance of keycloak is spun up only for testing?

met-api/src/met_api/services/widget_timeline_service.py
VineetBala-AOT: We have intentionally designed our models to be minimalist, focusing on retrieving requested data. Not all models support full CRUD operations; for instance, tables like submissions are restricted from updates, and deletion is limited, typically reserved for entities like widgets. All business logic is centralized in the service layer. 
Moreover the logic in timeline was causing a cyclic dependency because we had the timeline event service being imported back to the model so I had to move that logic to the service.

met-web/tests/unit/components/engagement/form/edit/EngagementForm.Edit.Two.test.tsx
VineetBala-AOT: Its just checking if the element exists.

PR #2360, opened 2024-01-18
DESENG-481: Adding unit tests for met api

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-481

*Description of changes:*
- Added new unit tests for met api


met-api/src/met_api/services/engagement_service.py
```diff
@@ -56,7 +56,8 @@ def get_engagement(self, engagement_id) -> EngagementSchema:
 
             engagement = EngagementSchema().dump(engagement_model)
             engagement['banner_url'] = self.object_storage.get_url(engagement['banner_filename'])
-        return engagement
+            return engagement
+        return None
```
NatSquared, line 60: Why explicitly return `None`? This should not affect functionality; is this for legibility reasons?

met-api/src/met_api/utils/enums.py
```diff
@@ -21,6 +21,7 @@ class ContentType(Enum):
     JSON = 'application/json'
     FORM_URL_ENCODED = 'application/x-www-form-urlencoded'
     PDF = 'application/pdf'
+    FORM = 'application/x-www-form-urlencoded'
```
NatSquared, line 24: Why are we creating `FORM` with the same value as `FORM_URL_ENCODED`? Do we plan to change this later or can we reuse `FORM_URL_ENCODED`

VineetBala-AOT: Thanks for catching this. I have updated the code to use FORM_URL_ENCODED.

met-api/src/met_api/services/engagement_service.py
VineetBala-AOT: This is just for readability.

NatSquared (approved):
This PR looks good! Thanks for putting in the work to create all these test scenarios.

PR #2363, opened 2024-01-22
Deseng-463:Poll Widget: Back-end

ratheesh-aot:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-463

*Description of changes:*

1. Created Database models for Widget Poll, Poll Answers, Poll Response.
2. Created API to manage Widget Poll, Poll Answers, Poll Response.
3. Created Unit tests to test the code.


met-api/src/met_api/services/poll_answers_service.py
```diff
@@ -0,0 +1,30 @@
...
+    def create_bulk_poll_answers(poll_id: int, answers_data: list):
+        """Bulk insert of poll answers."""
+        try:
+            if len(answers_data) > 0:
+                PollAnswerModel.bulk_insert_answers(poll_id, answers_data)
+        except SQLAlchemyError as e:
+            raise BusinessException(str(e), HTTPStatus.INTERNAL_SERVER_ERROR) from e
+
+    @staticmethod
+    def delete_poll_answers(poll_id: int):
```
Baelx, line 28: If the PO thinks we should allow polls to be changed while they're being voted on, then we may need to do a check to see if answers have been responded to and prevent deletion if so.

met-api/src/met_api/utils/ip_util.py
```diff
@@ -0,0 +1,30 @@
+"""
+This module provides utility functions for handling IP addresses in a Flask application.
+
+It includes a function for hashing IP addresses using SHA256, ensuring secure and consistent
```
Baelx, line 4: Great comments at the top of this file!

met-api/src/met_api/utils/ip_util.py
```diff
@@ -0,0 +1,30 @@
...
+    Hashes the given IP address concatenated with the Flask secret key.
+
+    Args:
+    ip_address (str): The IP address to be hashed.
+
+    Returns:
+    str: The resulting SHA256 hash as a hexadecimal string.
+    """
+    # Retrieve the secret key from Flask configuration with a fallback empty string
+    secret_key = current_app.config.get('SECRET_KEY', '')
```
Baelx, line 27: We may want to change the naming of this environment variable to something more intuitive. Maybe `IP_HASH_KEY` ?

met-api/src/met_api/models/poll_responses.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @classmethod
+    def get_responses_by_participant_id(cls, poll_id, participant_id) -> list[PollResponse]:
+        """Get responses for a poll."""
+        return db.session.query(PollResponse).filter(PollResponse.poll_id == poll_id,
+                                                     PollResponse.participant_id == participant_id,
+                                                     PollResponse.is_deleted == false()).all()
+
+    @classmethod
+    def update_response(cls, response_id, response_data: dict) -> PollResponse:
```
Baelx, line 39: I don't think there will be a need to update user responses. Once a user votes, they can't change it. I don't think we need this method

met-api/src/met_api/utils/ip_util.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # Concatenate the IP address and secret key, and hash the resulting string
+    return sha256(f'{ip_address}{secret_key}'.encode()).hexdigest()
```
Baelx, line 30: This hashes the IP, but where is the un-hash or IP comparison method? Is that going to come later?

Baelx (commented):
Great work so far! I'd like to discuss a few things before proceeding:
- Can you let me know how the IP comparison is happening? I'm not seeing where the IP address of the current response is being compared to all other unhashed IP fragments. Is that functionality coming in a future PR?
- I don't think we need a method to be able to update poll responses. Let me know if I'm missing something
- I just want to make sure we're only allowing one response from a user for a given poll ID, not 10. I may be missing something though

Baelx (approved):

ratheesh-aot:
Thank you for the review, @Baelx .

Quick summary based on our discussion: We're hashing IPs with our secret key and matching them against the database to flag any hash polled over 10 times. This threshold accounts for shared IPs in a network. We decided on the number 10 based on the requirement from the [ticket](https://apps.itsm.gov.bc.ca/jira/browse/DESENG-463).

We plan to add an 'is_deleted' metadata field to poll responses for better handling of suspicious entries, as per the[ epic ticket's](https://apps.itsm.gov.bc.ca/jira/browse/DESENG-455) requirements and thats why we had the update function to Poll Response.

I've corrected the typos you mentioned and included the missing code for IP fragmentation before hashing. The SECRET_KEY we're using is from the existing Flask configuration, not a new variable.

Regarding the logic to prevent admin users from editing/removing poll answers post-publication, I'll include that in the upcoming PR for the Poll widget UI. This was confirmed post the current PR.

Appreciate your feedback and support.

PR #2376, opened 2024-02-07
DESENG-447 : Convert keycloak groups to composite roles for permission levels

ratheesh-aot:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-447

*Description of changes:*

1. Commented out unit test related to Keycloak groups
2. Remove reference of Keycloak groups to roles
3. Commented out code related to Keycloak groups


met-api/tests/unit/api/test_user_membership.py
```diff
@@ -14,19 +14,19 @@
 
 """Tests to verify the user membership operations.
 
-Test-Suite to ensure that the user membership endpoints are working as expected.
+Test-Suite to ensure that the user membership endpoints are working as expected. # noqa: E501
```
Baelx, line 17: What does this mean here?

met-web/src/components/userManagement/userDetails/AddToEngagement.tsx
```diff
@@ -36,6 +36,7 @@ import { HTTP_STATUS_CODES } from 'constants/httpResponseCodes';
 export const AddToEngagementModal = () => {
     const { savedUser, addUserModalOpen, setAddUserModalOpen, getUserMemberships, getUserDetails } =
         useContext(UserDetailsContext);
+    console.log('user', savedUser);
```
Baelx, line 39: We can remove the logging for now. I should mentioned some calls were still present in the code

Baelx (changes requested):
Things look good! Thanks for doing this.
Could you please just remove logging like `print()` and `console.log()` if they're not needed. I had put them there for testing but we can remove them if they're not serving any other purpose

met-api/tests/unit/api/test_user_membership.py
ratheesh-aot: some codes and docstrings exceeded the max length when we commented it out that's why there is a ` # noqa: E501 ` to ignore that flake8 warning for the time being as we are gonna uncomment those functions in the near future

met-web/src/components/userManagement/userDetails/AddToEngagement.tsx
ratheesh-aot: removed

Baelx (approved):

PR #2377, opened 2024-02-08
DESENG-484: Adding max age for cors

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-484

*Description of changes:*
    - Introduces a new configuration variable to specify the maximum age for Cross-Origin Resource Sharing (CORS)
    - Modified the CORS preflight method to utilize this newly introduced variable.

### DIFF
--- a/met-api/src/met_api/config.py
+++ b/met-api/src/met_api/config.py
@@ -163,6 +163,12 @@ class Config:  # pylint: disable=too-few-public-methods
     # CORS settings
     CORS_ORIGINS = os.getenv('CORS_ORIGINS', '').split(',')
 
+    # CORS_MAX_AGE defines the maximum age (in seconds) for Cross-Origin Resource Sharing (CORS) settings.
+    # This value is used to indicate how long the results of a preflight request (OPTIONS) can be cached
+    # by the client, reducing the frequency of preflight requests for the specified HTTP methods.
+    # Adjust this value based on security considerations.
+    CORS_MAX_AGE = os.getenv('CORS_MAX_AGE', None) # Default: 0 seconds
+
     EPIC_CONFIG = {
         'ENABLED': env_truthy('EPIC_INTEGRATION_ENABLED'),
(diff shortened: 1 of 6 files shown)


met-api/src/met_api/config.py
```diff
@@ -163,6 +163,12 @@ def SQLALCHEMY_DATABASE_URI(self) -> str:
     # CORS settings
     CORS_ORIGINS = os.getenv('CORS_ORIGINS', '').split(',')
 
+    # CORS_MAX_AGE defines the maximum age (in seconds) for Cross-Origin Resource Sharing (CORS) settings.
+    # This value is used to indicate how long the results of a preflight request (OPTIONS) can be cached
+    # by the client, reducing the frequency of preflight requests for the specified HTTP methods.
+    # Adjust this value based on security considerations.
+    CORS_MAX_AGE = os.getenv('CORS_MAX_AGE', None) # Default: 0 seconds
```
Baelx, line 170: I like that you did this with an environment variable! A couple things:
- Could we make sure to set this in Openshift before we close the ticket? Looks like 2 hours is the maximum value for modern chromium browsers (Firefox is 72 hrs). Either this or we could set the default to be 2 hours.
- Our server is CORS-aware and checking for CORS origins so what are the security considerations here?

Baelx (changes requested):
Nice work ! Looks like we may be limited in truly reducing OPTIONS calls but the max age value will hopefully help a lot!

Could you please review my question before we proceed with his PR?

met-api/src/met_api/config.py
VineetBala-AOT: Sure, I will set it in Openshift before closing the ticket.

For enhanced security considerations, configuring Access-Control-Max-Age as an environment variable provides the flexibility to adjust its duration or even eliminate it entirely in response to security concerns. Online recommendations suggest maintaining a shorter duration for Access-Control-Max-Age to mitigate potential risks.

Baelx (approved):

PR #2399, opened 2024-02-28
[ TO FEATURE ] DESENG-509 Creating model and migration for Language table

ratheesh-aot:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-509

*Description of changes:*
Creating model and migration for Language table


met-api/src/met_api/models/language.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Language(BaseModel):
+    """Definition of the Language entity."""
+
+    __tablename__ = 'language'
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String(50), nullable=False)  # eg. English, French etc
+    code = db.Column(db.String(2), nullable=False)  # eg. en, fr etc
```
Baelx, line 17:
```suggestion
    code = db.Column(db.String(2), nullable=False, unique=True)  # eg. en, fr etc
```

We want to make sure that the language `code` is unique, as per the ticket. For example, we don't want users to be able to enter "fr" twice. Is there a better way to do this other than specifying it in the DB model?

Baelx (changes requested):
Looks good, please just review my comment about the `code` column being unique.

Baelx (approved):
Thanks for implementing my feedback!

PR #2417, opened 2024-03-14
[To Feature] DESENG-467 Enable string translations for public view static text

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-467

*Description of changes:*
- Implemented a language selector in the public header.
- Incorporated logic to dynamically adjust the unauthenticated route based on the selected language and load the appropriate translation file.
- Enhanced all public pages to fetch static text from the translation file.
- Added sample files for French language.


Baelx:
As discussed with Vineet, we're going to wait until the PO returns on Monday to see if this PR meets the requirements of the ticket. Please don't review or merge until then.

met-web/src/components/common/LanguageSelector.tsx
```diff
@@ -0,0 +1,105 @@
...
+                }),
+            );
+
+            // Change the URL when the language is changed
+            const pathSegments = location.pathname.split('/');
+            const languageIndex = pathSegments.findIndex((seg) => seg.length === 2); // Find index of the 2-character language code
+            if (languageIndex !== -1) {
+                pathSegments[languageIndex] = selectedLanguage;
+            } else {
+                // Language code not found in path, insert it at the appropriate position
```
Baelx, line 41: Thanks for these helpful comments!

met-web/src/components/common/LanguageSelector.tsx
```diff
@@ -0,0 +1,105 @@
...
+                pathSegments.splice(1, 0, selectedLanguage);
+            }
+            // navigate(pathSegments.join('/'));
```
Baelx, line 44: I'm guessing this needs to be removed

met-web/src/components/layout/Footer/index.tsx
```diff
@@ -142,7 +143,7 @@ const Footer = () => {
                     </Stack>
                 </Grid>
                 <Grid item xs={12} sm={6} container justifyContent={'flex-end'} alignItems="flex-end">
-                    <MetBody>© 2023 Government of British Columbia</MetBody>
+                    <MetBody>{translate('footer.copyrightNotice')}</MetBody>
                 </Grid>
             </Grid>
         </Grid>
```
Baelx, line 149: There's an `alt` tag in this file that still requires the `translate` function. Any text that could possibly appear to the user on the front-end will need to be translatable. This includes `alt`, `aria-label`, `placeholder`, etc.

met-web/src/components/publicDashboard/Dashboard.tsx
```diff
@@ -52,7 +60,7 @@ const Dashboard = () => {
             })
             .catch((error) => {
                 // Error handling
-                console.error('Error generating dashboard PDF:', error);
+                console.error(translate('dashboard.consoleError.pdfError'), error);
```
Baelx, line 63: Console errors don't need translation so no need to handle these. We're making the assumption that our team's developers work in English

Baelx (changes requested):
Thanks! Just before approving, could you please remove the commented-out code if it's not needed and then correct one more `alt` tag that still needs to use `translate`

met-web/src/components/common/LanguageSelector.tsx
VineetBala-AOT: Removed thanks.

met-web/src/components/layout/Footer/index.tsx
VineetBala-AOT: Updated the alt tag, thanks

met-web/src/components/publicDashboard/Dashboard.tsx
VineetBala-AOT: Removed the translation for console error. Thanks.

Baelx (approved):

PR #2426, opened 2024-03-21
[To Feature] DESENG-501 Removing content from engagement table

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-501

*Description of changes:*
- Backend changes
- Removed content and rich content from engagement tables as these will be part of the new tables for engagement content.
- Added logic to create default summary content when an engagement in created.
- Added logic to create a default custom content content when a custom tab is added.
- Related changes to unit test to remove content and rich content from engagement tables.


met-api/src/met_api/services/engagement_content_service.py
```diff
@@ -36,10 +38,29 @@ def create_engagement_content(engagement_content_data, engagement_id):
...
 
+    @staticmethod
+    def create_default_custom_content(eng_id: int, eng_content_id: int):
+        """Create default engagement custom content."""
+        default_summary_content = {
+            'engagement_id': eng_id
+        }
+        try:
+            EngagementCustomContentService.create_custom_content(eng_content_id, default_summary_content)
+        except Exception as exc:  # noqa: B902
```
Baelx, line 58: What does "# noqa: B902" mean here?

Baelx (approved):

met-api/src/met_api/services/engagement_content_service.py
VineetBala-AOT: The linter suggests adding specific exceptions, but I've opted for a general exception handling approach to ensure that any error encountered during the creation of custom content raises a business exception error.

PR #2443, opened 2024-04-04
[ To Main ] - DESENG-508 - Updated Keycloak unit test for new CSS API

ratheesh-aot:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-508

*Description of changes:*
Updated Keycloak Unit Tests for New CSS API Integration.

### DIFF
--- a/met-api/src/met_api/services/keycloak.py
+++ b/met-api/src/met_api/services/keycloak.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 """Utils for keycloak administration."""
 
-import json
 from typing import List
 
 import requests
--- a/met-api/src/met_api/services/keycloak.py
+++ b/met-api/src/met_api/services/keycloak.py
@@ -67,6 +67,7 @@ class KeycloakService:  # pylint: disable=too-few-public-methods
         for user_id in user_ids:
             query_user_url = (f'{self.base_url}/{self.integration_id}/'
                               f'{self.environment}/users/{user_id}/roles')
+            print(query_user_url)
             response = requests.get(query_user_url, headers=headers, timeout=self.timeout)
             if response.status_code == 200:
                 if (roles := response.json().get('data')) is not None:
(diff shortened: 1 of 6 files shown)


Baelx:
@ratheesh-aot So given the recent development that we won't be using the CSS API anymore, do you think any of this work can be preserved for later? 😓

ratheesh-aot:
@Baelx Some API tests can be useful again. Also to make sure all tests are working until we start on the new Role ticket and to make sure some redundant "add to group " functionalities are removed, we may merge this PR for now. And update it later

met-api/tests/unit/services/test_keycloak.py
```diff
@@ -16,38 +16,135 @@
...
+from unittest.mock import patch
+
+
+KEYCLOAK_GET_REQUEST = 'met_api.services.keycloak.requests.get'
+KEYCLOAK_POST_REQUEST = 'met_api.services.keycloak.requests.post'
+KEYCLOAK_DEL_REQUEST = 'met_api.services.keycloak.requests.delete'
+KEYCLOAK_SERVICE = KeycloakService()
+
+
+@pytest.fixture()
```
NatSquared, line 31: Is this patch intended to run before each function? If so, maybe explicitly specify `fixture(scope='function')`. If not, specify another scope such as `module`

met-api/src/met_api/services/keycloak.py
```diff
@@ -76,31 +75,6 @@ def get_users_roles(self, user_ids: List):
 
         return user_role_mapping
 
-    # @staticmethod
```
NatSquared, line 79: thanks for cleaning up all these comments!

NatSquared (approved):
These are some really high-quality tests! Thank you :3

met-api/tests/unit/services/test_keycloak.py
ratheesh-aot: @NatSquared  Thanks for notifying this. When I checked the default scope of the fixture will be **function** always. We need that because we have a test named def **test_get_admin_token**(mock_post) that checks whether _get_admin_token is called once or not. Also, all other test functions require this patch. That is why I left it to default scope which is function.

PR #2456, opened 2024-04-10
DESENG-542: Keycloak auth method and redirection fixed

ratheesh-aot:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-542

*Description of changes:*
 - Created authentication provider to know the logged-in state before the App component is rendered.
 - Optimized tenant identification from the url.
 - Optimized path segment identification.

Note: Keycloak-js has an [issue ](https://www.keycloak.org/docs/latest/securing_apps/#_modern_browsers)with Firefox (Tracking Protection) in local machines (or non-https) . So if using the met-web app locally, please turn off the tracking protection.

### DIFF
--- a/met-web/src/utils/index.ts
+++ b/met-web/src/utils/index.ts
@@ -56,7 +56,7 @@ export const determinePathSegments = () => {
 };
 
 export const findTenantInPath = () => {
-    // filters out empty segments, which can occur if there are leading or trailing slashes in the pathname
+    // finding tenant from the path segments
     const pathSegments = determinePathSegments();
     return pathSegments.length > 0 ? pathSegments[0].toLowerCase() : '';
 };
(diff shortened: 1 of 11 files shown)


met-web/src/utils/index.ts
```diff
@@ -47,3 +47,16 @@ export const blobToFile = (theBlob: Blob, fileName: string): File => {
...
 };
+
+export const determinePathSegments = () => {
+    const url = new URL(window.location.href);
+    // filters out empty segments, which can occur if there are leading or trailing slashes in the pathname
+    const pathSegments = url.pathname.split('/').filter((segment) => segment.trim() !== '');
+    return pathSegments;
+};
+
+export const determineBaseName = () => {
```
Baelx, line 58: I think the name of this function is potentially a bit confusing. When I think of a URL's "base" I would think of the domain name itself. Here it looks like this function is meant to find the tenant code at the beginning of the URL path?

Maybe you could change the comment to explain the function a bit more?

Baelx (approved):
Everything looks good pretty much. Before merging, could I ask you to maybe increase the clarity around the `determineBaseName` function?

met-web/src/utils/index.ts
ratheesh-aot: Make sense! Updated

PR #2470, opened 2024-04-22
[To Main] In-app role management

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-574

*Description of changes:*
    - Introduced new database models to manage user roles and user-to-role mappings in the Met database.
    - Revamped the authentication validation process to rely on roles retrieved directly from the database,
    rather than relying solely on token-based roles.
    - Eliminated redundant or unused roles from the system.
    - Updated unit tests to to fetch roles directly from the database instead of relying on role generation from the factory and eliminated the Keycloak mock.
    - Renamed admin and viewer role.
    - Removed the usage of CSS API for role management.


Baelx (approved):
Great work on this! Approved

PR #2472, opened 2024-04-25
[To feature] Upgrading the version of python on met analytics

VineetBala-AOT:

*Description of changes:*
- Modify code as needed if the newer version does not support the former.


analytics-api/tests/conftest.py
```diff
@@ -82,7 +82,7 @@ def session(app, db):  # pylint: disable=redefined-outer-name, invalid-name
         txn = conn.begin()
 
         options = dict(bind=conn, binds={})
-        sess = db.create_scoped_session(options=options)
+        sess = db._make_scoped_session(options=options)
```
Baelx, line 85: It looks like they made this method private. I'm curious, is there a different one that developers need to use to create a scoped session?

Baelx (approved):

analytics-api/tests/conftest.py
VineetBala-AOT: Updated it, thanks for noting this.

PR #2473, opened 2024-04-25
[To feature] Upgrading the version of python on notify-api

VineetBala-AOT:

*Description of changes:*
- Modify code as needed if the newer version does not support the former.


notify-api/src/notify_api/resources/sms.py
```diff
@@ -32,4 +32,5 @@ def post():
         """Send notification."""
         # sms_payload = request.get_json(force=True)
         # get_sms_service().send(sms_payload)
+        # pylint: disable=W0719
```
Baelx, line 35: For each of these unimplemented methods, could we add a TODO for us to update them and remove the generic `Exception` at some point?

Baelx (approved):
Approved! Nice work.

It looks like generic `Exception` calls are discouraged in Python but I understand they're just to signal unimplemented methods. Maybe a TODO in each of those would remind developers to remove the Exception when it's time to implement those.

notify-api/src/notify_api/resources/sms.py
VineetBala-AOT: Added TODO as suggested. Thanks.

PR #2474, opened 2024-04-25
[To feature] Upgrading the version of python on met api

VineetBala-AOT:

*Description of changes:*
- Modify code as needed if the newer version does not support the former or lint errors.


met-api/src/met_api/services/cdogs_api_service.py
```diff
@@ -54,6 +54,7 @@ def generate_document(self, template_hash_code: str, data, options):
 
     @staticmethod
     def _post_generate_document(json_request_body, headers, url):
+        # pylint: disable=missing-timeout
         response = requests.post(url, data=json_request_body, headers=headers)
```
Baelx, line 58: Was there a reason we couldn't remedy this linter issue? Even if we don't want to specify a timeout value, couldn't we just set it to None?

```suggestion
        response = requests.post(url, data=json_request_body, headers=headers, timeout=None)
```

met-api/src/met_api/services/poll_response_service.py
```diff
@@ -83,6 +84,7 @@ def get_poll_details_with_response_counts(poll_id):
             db.session.query(
                 PollAnswer.id.label('answer_id'),
                 PollAnswer.answer_text,
+                # pylint: disable=not-callable
                 func.count(PollResponseModel.selected_answer_id).label('response_count')
```
Baelx, line 88: From what I've read, it's a known issue for pylint to be labelling `func` method calls as false positives for this rule. You may be able to get around the issue with this stackoverflow answer: https://stackoverflow.com/a/75789407

Baelx (approved):
Approved! You may just want to review some ways to correct those linter errors instead of putting in exceptions, if possible.

met-api/src/met_api/services/cdogs_api_service.py
VineetBala-AOT: Updated as suggested, thanks.

PR #2489, opened 2024-05-06
[To Main] Upgrading the version of met-formio

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-596

*Description of changes:*
- Upgrading the version of met-formio to add a default description for the form components.

### DIFF
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -1,7 +1,8 @@
 ## May 6, 2024
 
 - **Feature** Auto-populate question descriptions [🎟️ DESENG-596](https://apps.itsm.gov.bc.ca/jira/browse/DESENG-596)
-  - Upgrading the version of met-formio.
+  - Upgrading the version of met-formio to version 1.0.15-rc1. This version has the changes to have a default 
+  description for the form components. The changes were pushed to met-formio by the PR: https://github.com/bcgov/met-formio/pull/25.
 
 ## May 2, 2024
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.MD
```diff
@@ -1,3 +1,8 @@
+## May 6, 2024
+
+- **Feature** Auto-populate question descriptions [🎟️ DESENG-596](https://apps.itsm.gov.bc.ca/jira/browse/DESENG-596)
```
Baelx, line 3: From the Changelog, and by merging into Main, it looks like this feature is complete just by upgrading the version. Is that the case?

Baelx (approved):
To me, it's not super clear if this feature is done or not. It may create confusion for folks looking in the history. 

If you're wanting to do some preparatory work for this ticket by first upgrading the library in the main branch, I'd change the Changelog entry to read something like:

`- Prepare for DESENG-596 by upgrading the met-formio library`

This would make it clear that the feature still needs to be completed but that you're doing work to prepare for it.

You're welcome to merge in any case.

CHANGELOG.MD
VineetBala-AOT: Yes the changes were pushed to met-formio and published to npm. This version of met-formio has the changes needed for this ticket. I update the changelog to have more details.

Baelx: Ohhh okay so the substantive changes were in the met-formio repository for this. Thanks for clarifying

PR #2495, opened 2024-05-07
[To Main] DESENG-478 - Override _add_tenant_filter to check in user group membership table for tenant mapping

ratheesh-aot:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-478

*Description of changes:*
 - Tested users can belong to more than one tenant.
  - Override the default tenant filter for the **staff_user** table to check with the **user_group_membership** table.
  
  Note : tenant_id is left as it is in the staff_user table to track the tenant_id of the user when it is created first time. 


Baelx (approved):
Please just remove the print statement before merging!

PR #2496, opened 2024-05-08
Bugfix/deseng575 Language picker should only appear on engagements with more than one language

Baelx:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-575

*Description of changes:*
- Language picker should only appear on engagements with more than one language
- Language picker disappears completely when not viewing an engagement
- Allow for the web app to see if there are available translations by language
- Disable language switching itself for now until design decisions are made

I've had to kneecap the language switcher functionality until we can make some design decisions regarding what to do about:
- Language switching for engagements only vs. whole-site
- Sites that have differing languages between static and engagement content
- Location of the language switcher (we may move it so it's nested in the engagement rather than in the header)


met-web/src/components/engagement/view/EngagementView.tsx
```diff
@@ -21,6 +23,8 @@ export const EngagementView = () => {
     const isLoggedIn = useAppSelector((state) => state.user.authentication.authenticated);
     const isPreview = isLoggedIn;
     const { savedEngagement } = useContext(ActionContext);
+    const { setEngagementViewMounted, setAvailableEngagementTranslations } = useContext(LanguageContext);
+    const [test, setTest] = useState(false);
```
Baelx, line 27: Will remove this

ratheesh-aot (approved):
LGTM!

Baelx:
> LGTM!

Thanks @ratheesh-aot . I'm just finishing fixing the linter/test errors and will merge in a second

PR #2501, opened 2024-05-09
[To Main] Adding a new no results page

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-586

*Description of changes:*
  - Added a new no results page to display to users when their search term(s) and / or filter(s) returns no engagement results on landing page.
  - Contact information is developed but hidden currently as this was the approach requested in the ticket till the contact information is decided.
  - Added new unit test.


met-web/src/routes/NoResults.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    direction="column"
+                    alignItems="center"
+                    padding={0}
+                    width="100%"
+                    maxWidth="1120px"
+                    height="auto"
+                    border="1px solid #FAF9F8"
+                    boxShadow="0px 0.6px 1.8px rgba(0, 0, 0, 0.1), 0px 3.2px 7.2px rgba(0, 0, 0, 0.13)"
+                    borderRadius="8px"
+                    display={'none'} // TODO: remove this once contact information is decided
```
Baelx, line 59: I appreciate that you want to save us some work with the contact area, but I'm hesitant to add this code because of the uncertainty. If we decide not to have a contact area, we'll have to go back and remove all this.

Did you speak to Jess or the PO about this? Did they instruct you to add it anyway? If we don't have explicit direction from the PO on this, then I'd recommend we remove it.

Baelx (changes requested):
Things look good except for the contact information. I really appreciate that you wanted to add this to save us work but I think there's a high chance that we will never use this functionality and as such, I don't think we should have it in the code.

Unless Steve explicitly asked for us to include it so we can figure it out later, I'd say we could keep it. But from the ticket, and my experience with the Citizen Engagement team, I'd say that we should err on the side of not having it.

Please let me know if you had a discussion with Steve about this or if you'll be removing the contact info section.

met-web/src/locales/en/gdx.json
```diff
@@ -309,5 +309,26 @@
             "The page you're looking for cannot be found.",
             "The page you're looking for might have been removed, moved or is temporarily unavailable."
         ]
+    },
+    "NoResults": {
```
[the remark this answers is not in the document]
VineetBala-AOT: I think until we work on the ticket DESENG-611 which is to have the static text based on language instead of tenant we will have to maintain in both places else it will not display the text in GDX tenant.

met-web/src/routes/NoResults.tsx
VineetBala-AOT: I did talk to Jess about it and she suggested to add the contact and make it hidden as is mentioned in the ticket as well.

met-web/src/locales/en/gdx.json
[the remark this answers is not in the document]
Baelx: Ah okay so what I understand from this is that there's no way to mix locale files. In other words, if there's a gdx.json locale file, none of the default.json locale strings can be used. Thus, we need to duplicate every string for now, is that right?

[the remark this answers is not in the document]
VineetBala-AOT: Yes that's correct.

met-web/src/routes/NoResults.tsx
Baelx: Thanks for confirming you spoke with someone about this. This feels like one of those situations where we, as developers, need to decide what's best for the code. Technically, the PO can let us know what's needed for the user, but how we do that is up to us. From a maintainability standpoint, my instinct is usually not to build things on the chance that we use them again, but to build as needed.

That said, it's not that much code to leave in, and hopefully the PO and the designer have a plan for it in the short term. We can proceed with it.

Baelx (approved):
Thanks for getting back to me on my questions. You can proceed with merge!

met-web/src/routes/NoResults.tsx
VineetBala-AOT: Thanks will follow this going forward.

PR #2503, opened 2024-05-09
[To Main] Add tenant listing page

NatSquared:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-592

*Description of changes:*
- **Feature** Added the initial version of the tenant listing page
  - The page lists all tenants in the system
  - Functionality to view or create individual tenants is still under construction
  - Began work on adding new components from the MET component library, with high reusability in mind


met-web/src/components/MetDesignSystem/Navigation/Breadcrumb.tsx
```diff
@@ -0,0 +1,29 @@
...
+    name: string;
+    link?: string;
+};
+
+export const BreadcrumbTrail: React.FC<{ crumbs: BreadcrumbProps[]; smallScreenOnly?: boolean }> = ({
+    crumbs,
+    smallScreenOnly,
+}) => {
+    return (
+        <Breadcrumbs aria-label="breadcrumb" sx={smallScreenOnly ? { display: { xs: 'block', md: 'none' } } : {}}>
```
Baelx, line 15: I feel like it'd be a bit more Reactive to have screen size check happen outside the component and have it not render, rather than having it built in and trigger a display:none. Although I get the impulse to have a component be super self-contained.

met-web/src/components/MetDesignSystem/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+};
+
+export const globalFocusShadow = `inset 0px 0px 0px 2px ${colors.focus.regular.inner}`;
+
+export const globalFocusVisible = {
+    '&:focus-visible': {
+        outline: `2px solid ${colors.focus.regular.outer}`,
+        boxShadow: globalFocusShadow,
+    },
+};
```
Baelx, line 105: I appreciate that you took the trouble to get all these styling variables accounted for - and for creating those new design system components.

MET already has lots of common components in use. Just curious if you looked into how much of it we can massage into the Design System rather than creating everything from scratch.

met-web/src/components/tenantManagement/Listing.tsx
```diff
@@ -0,0 +1,114 @@
+import { AddCircleOutline } from '@mui/icons-material';
```
Baelx, line 1: Should this be with fontawesome?

met-web/src/components/tenantManagement/Listing.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { Else, If, Then } from 'react-if';
+import { BreadcrumbTrail } from 'components/MetDesignSystem/Navigation/Breadcrumb';
+
+const TenantListing = () => {
+    const [tenants, setTenants] = React.useState<Tenant[]>([]);
+    const [loading, setLoading] = React.useState<boolean>(true);
+
+    useEffect(() => {
+        const fetchTenants = async () => {
+            const tenants = await getAllTenants();
```
Baelx, line 28: You might be able to save a couple lines of code(plus add more specific error handling), and the function declaration with:

```suggestion
            getAllTenants
            .then(returnedTenants => setTenants(returnedTenants)
            .catch(// Handle Error)
            .finally(() => setLoading(false))
```

met-web/src/components/tenantManagement/Listing.tsx
```diff
@@ -0,0 +1,114 @@
...
+                                        </TableCell>
+                                        <TableCell>
+                                            <Skeleton variant="text" width={480} />
+                                        </TableCell>
+                                    </TableRow>
+                                ))}
+                            </Then>
+                            <Else>
+                                {tenants.map((tenant) => (
+                                    <TableRow onClick={() => console.log(tenant.name)} key={tenant.name} tabIndex={0}>
```
Baelx, line 92: console.log left in

Baelx (changes requested):
Looks great! Could you please:
- remove the console.log
- review the naming of "Feedback" in resources/tenant.py
- review if the material UI icon can be used with fontawesome (or let me know if it's not possible)

I'd also love to just chat briefly about some of the infrastructure decisions you made with the new design system components. It would be great if all the developers were on the same page with how we're going to start integrating Jess' design system :)

Baelx (approved):
I'm not in the office today so I'll switch my review to Approved so you're not blocked!

met-web/src/components/tenantManagement/Listing.tsx
NatSquared: Unfortunately, `.finally()` isn't available because we're [compiling in ES2017 mode](https://github.com/bcgov/met-public/blob/fc97bcb2481bdf39b9b795172d2254efb8b74303/met-web/tsconfig.json#L14)

met-web/src/components/tenantManagement/Listing.tsx
NatSquared: Yes; I was waiting for the Font Awesome PR to be merged

Baelx: Ahh yes, sorry

met-web/src/components/tenantManagement/Listing.tsx
Baelx: I believe you should be able to use it if you want to still. I think the `target` option just indicates which ES features are "downgraded" and which are left intact. 

https://www.typescriptlang.org/tsconfig/#target

Baelx (approved):
Thanks! Approved

PR #2506, opened 2024-05-13
[To Main] Add metadata management role

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-603

*Description of changes:*
- Implemented a new role named "manage_metadata" within the Admin group to restrci access for metadata management
- Updated the frontend to restrict access to the "metadata management" link in the menu for users without the newly added role
- Backend changes to incorporate the new role for access control purposes, ensuring only authorized users can perform metadata management actions


met-api/src/met_api/resources/engagement_metadata.py
```diff
@@ -75,9 +73,16 @@ class EngagementMetadata(Resource):
     @cross_origin(origins=allowedorigins())
     @API.doc(security='apikey')
     @API.marshal_list_with(metadata_return_model)
-    @require_role(VIEW_ENGAGEMENT_ROLES)
+    @_jwt.requires_auth
     def get(engagement_id):
         """Fetch engagement metadata entries by engagement id."""
+        authorization.check_auth(
```
Baelx, line 79: Levels of abstraction are tricky here, but the ticket still calls for engagement metadata to allow CRUD operations for users who can edit engagements.

I believe it's the metadata_taxon file that will restrict its operations to the `MANAGE_METADATA` role.

Baelx (changes requested):
I haven't completed review of the tests yet but could you please update the changes to allow users who could edit engagement metadata to be able to perform CRUD operations on them? We want to make sure that many users can still access/modify metadata on individual engagements. Taxons should be restricted to metadata managers.

Please see the ticket or speak with Nat for more information!

met-api/src/met_api/resources/engagement_metadata.py
VineetBala-AOT: Thanks, changed it back to use edit engagement. The only change here being allowing team members assigned to the engagement to edit as well.

met-api/src/met_api/resources/engagement_metadata.py
```diff
@@ -131,11 +146,16 @@ class EngagementMetadataById(Resource):
 
     @staticmethod
     @cross_origin(origins=allowedorigins())
-    @require_role(VIEW_ENGAGEMENT_ROLES)
+    @_jwt.requires_auth
```
Baelx, line 149: Are we sure about requiring authorization for this route? What if we're pulling engagement metadata on the frontend for public users?

Baelx (approved):
I chatted with Vineet about my questions:
- We don't need to allow public access to metadata for engagements yet. Tickets are coming for how we handle metadata on the frontend
- Tests look for user groups which include the new metadata manager role, so we're covered in terms of testing it

PR #2507, opened 2024-05-14
[To Main] DESENG-579: Add language to survey email links

NatSquared:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-579

*Description of changes:*
- **Bugfix** Add language code to email links
  - Added a language code to email template links so that they point to a valid page again
  - The system tries to determine the language from the user's browser, but if it can't, it defaults to English
  - A future, larger refactor could allow us to remember the user's language preference and always use that


met-api/src/met_api/services/email_verification_service.py
```diff
@@ -186,27 +189,26 @@ def _render_subscribe_email_template(survey: SurveyModel, token, subscription_ty
         return subject, body, args, template_id
 
     @staticmethod
-    def _render_survey_email_template(survey: SurveyModel, token):
+    def _render_survey_email_template(survey: SurveyModel, token, lang_code):
         # url is origin url excluding context path
         engagement: EngagementModel = EngagementModel.find_by_id(survey.engagement_id)
-        engagement_name = engagement.name
```
Baelx, line 192: I can see why you removed this. Not sure why we'd need a dedicated variable for it.

Baelx (approved):
LGTM!

PR #2509, opened 2024-05-16
[To Main] Replace Material Icons with Font Awesome

VineetBala-AOT:
Issue #: https://apps.itsm.gov.bc.ca/jira/browse/DESENG-608

*Description of changes:*
  - Replaced Material Icons with Font Awesome for all icons for which we had a mapping
  - Noted a few icons for which we did not have a mapping and those have been added to the ticket

### DIFF
--- a/met-web/src/components/metadataManagement/TaxonEditor.tsx
+++ b/met-web/src/components/metadataManagement/TaxonEditor.tsx
@@ -1,7 +1,11 @@
 import { Grid, Box, Paper, IconButton, Modal, Button, Typography, Chip, Fade } from '@mui/material';
 import { useTheme } from '@mui/material/styles';
 import useMediaQuery from '@mui/material/useMediaQuery';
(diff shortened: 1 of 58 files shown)


met-web/src/components/metadataManagement/TaxonEditor.tsx
```diff
@@ -273,11 +277,12 @@ export const TaxonEditor = () => {
                             }}
                         >
                             <Chip
+                                icon={<FontAwesomeIcon icon={faChevronDown} />}
                                 variant="filled"
                                 color="secondary"
                                 label="Scroll for more"
-                                icon={<KeyboardArrowDown />}
                                 onClick={() => {
+                                    // Your click handler function
```
Baelx, line 285: This could probably be removed

Baelx (approved):
I would say just remove the comment that doesn't add anything, see my note. But other than that you can go ahead and merge!

met-web/src/components/metadataManagement/TaxonEditor.tsx
VineetBala-AOT: Removed, thanks

PR #2533, opened 2024-06-04
DESENG-632 Add DB tables & API routes

Baelx:
Issue #: https://github.com/bcgov/met-public/issues/

*Description of changes:*
- Add DB tables, models
- Add API routes
- Update Readme for migrations folder

This is a partial PR.
- Tests will come later, as will securing API routes
- Changelog will be updated when all is merged into main


met-api/migrations/versions/22fb6b5b5aed_relax_language_table_constraints.py
```diff
@@ -0,0 +1,27 @@
+"""
+Relax language table constraints.
```
NatSquared, line 2: like, just chill out, dude

met-api/migrations/versions/c656f3f82334_add_languages_and_tenant_mapping.py
```diff
@@ -0,0 +1,42 @@
...
+import json
+
+
+# revision identifiers, used by Alembic.
+revision = 'c656f3f82334'
+down_revision = '22fb6b5b5aed'
+branch_labels = None
+depends_on = None
+
+with open("migrations/supported_languages.json", "r") as read_file:
```
NatSquared, line 20: Has maintainability been considered here? What will it look like when we need to add or remove a language? Can we still use the same JSON file or will we be manually migrating to match any edits made?

met-api/src/met_api/resources/language.py
```diff
@@ -75,43 +76,82 @@ def delete(language_id):
...
-        """Create a new language."""
+    def get():
+        """Fetch list of languages."""
         try:
-            request_json = request.get_json()
-            valid_format, errors = schema_utils.validate(
-                request_json, 'language'
+            languages = LanguageService.get_languages()
+            return (
+                jsonify(LanguageSchema(many=True).dump(languages)),
```
NatSquared, line 153: Is this call to `jsonify` necessary if dump() already produces a json string?

met-api/src/met_api/resources/language.py
```diff
@@ -75,43 +76,82 @@ def delete(language_id):
...
+            ) from e
+
+
+@cors_preflight('DELETE, OPTIONS')
+@API.route('/mappings/<int:language_mapping_id>')
+class DeleteLanguageMapping(Resource):
+    """Resourse for removing language-tenant relationships."""
+
+    @staticmethod
+    # @_jwt.requires_auth
```
NatSquared, line 126: I realize this is temporary, but do remember to add appropriate authentication to your routes before merging to main

met-api/src/met_api/services/language_service.py
```diff
@@ -24,22 +23,6 @@ def get_languages():
         languages_records = Language.get_languages()
         return LanguageSchema(many=True).dump(languages_records)
 
-    @staticmethod
-    def create_language(language_data):
-        """Create language."""
```
NatSquared, line 29: If we can delete languages, we should probably still be able to add them as well, otherwise the number of languages can only decrease with time. Is there a reason all this code was removed?

NatSquared (approved):
Looks good! I would consider either:
- not removing the endpoints for adding a language
OR
- also removing the endpoints for deleting languages

met-api/migrations/versions/c656f3f82334_add_languages_and_tenant_mapping.py
Baelx: Great point!

Adding/removing languages in the future would be done with a DB migration. I don't expect us to have to do so that often. Given that, maybe it would be better to rename `supported_languages` to `initial_languages` or something, knowing that it may not be the exhaustive in the long term.

I could have also not had a separate file but I felt the size of the migration file would be super huge and less readable that way. Do you think maybe I should just move the list into the migration file given that they'll be the ultimate source of truth?

met-api/src/met_api/resources/language.py
Baelx: Oh I missed this, this might just solve an issue I was having... Thanks

met-api/src/met_api/resources/language.py
Baelx: I made a note about it in the PR description but yes for sure, I'll have to circle back to auth once I push changes for the frontend. It'll be easier to test it when I'm making requests from the browser

met-api/src/met_api/services/language_service.py
Baelx: If I left in a call to deleting languages themselves then I'll be removing it. Great point

met-api/src/met_api/resources/language.py
```diff
@@ -59,16 +60,63 @@ def patch(language_id):
...
+    # @cross_origin(origins=allowedorigins())
+    def delete(language_mapping_id):
+        """Remove a language mapping from a tenant."""
         try:
-            success = LanguageService.delete_language(language_id)
+            success = LanguageService.remove_language_mapping_from_tenant(language_mapping_id)
             if success:
-                return 'Successfully deleted language', HTTPStatus.NO_CONTENT
-            raise ValueError('Language not found')
+                return 'Successfully deleted language-tenant mapping', HTTPStatus.NO_CONTENT
```
NatSquared, line 118: I appreciate you updating the language to remain accurate :)

PR #2561, opened 2024-07-22
Feature/deseng657

jareth-whitney:
Issue #: https://citz-gdx.atlassian.net/browse/DESENG-657

- Small change to conditional in language selector due to an error on the test server. The conditional for checking if a language is right_to_left needs to account for the value not existing in the language entry.

### DIFF
--- a/met-web/src/components/engagement/new/create/index.tsx
+++ b/met-web/src/components/engagement/new/create/index.tsx
@@ -123,8 +123,9 @@ const EngagementCreationWizard = () => {
 
     const handleLanguageEntries = (languages: Language[]) => {
         return languages.map((language) => {
-            const languageWithDirection =
-                true === language.right_to_left ? language.name.split('').reverse().join('') : language.name;
+            const languageWithDirection = language.right_to_left
+                ? language.name.split('').reverse().join('')
+                : language.name;
             return (
                 <>
                     <span>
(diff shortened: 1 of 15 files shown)


met-web/src/components/engagement/new/create/index.tsx
```diff
@@ -123,8 +123,9 @@ const EngagementCreationWizard = () => {
...
     const handleLanguageEntries = (languages: Language[]) => {
         return languages.map((language) => {
-            const languageWithDirection =
-                true === language.right_to_left ? language.name.split('').reverse().join('') : language.name;
+            const languageWithDirection = language.right_to_left
+                ? language.name.split('').reverse().join('')
+                : language.name;
             return (
                 <>
                     <span>
```
NatSquared, line 131:
```suggestion
            return (
                <>
                    <span style={{ direction: language.right_to_left ? 'rtl' : 'ltr' }}>

NatSquared (approved):
Looks good - would recommend using CSS instead of reversing text

met-web/src/components/engagement/new/create/index.tsx
Baelx: Maybe I don't have the full context here but I don't believe we should implementing any logic to handle RTL languages, whether it's reversing a string or adding styling. If this is a simple field for selecting a language then I don't think engagement editors will expect the name of the language itself to show up reversed in some cases. 

The ticket doesn't make mention of RTL so I would check with the PO on this, or just omit language direction handling for now

PR #2565, opened 2024-07-24
feature/deseng661: Implemented new admin navigation design.

jareth-whitney:
Issue #: [🎟️ DESENG-661](https://citz-gdx.atlassian.net/browse/DESENG-661)

- Implemented new admin navigation menu
- This includes mobile and desktop viewports
- This does not include dark mode
- User and tenant sections will be handled in header ticket (as per Steve)


met-web/src/components/layout/SideNav/SideNav.tsx
```diff
@@ -28,40 +40,142 @@ const DrawerBox = () => {
...
+            Metadata: faTags,
+            Languages: faGlobe,
+            'User Admin': faUserGear,
+            'Tenant Admin': faHouseUser,
+            'MET Feedback': faMessagePen,
+        };
+        // If the route name doesn't match an expected page title, return the broken link icon
+        const possiblePageTitles: string[] = Object.keys(iconAssignments).filter((key) => key.includes(route.name));
+        if (0 === possiblePageTitles.length) {
+            return <FontAwesomeIcon icon={faLinkSlash} style={fontAwesomeStyles} />;
```
NatSquared, line 114: nice touch with the broken link icon :)

met-web/src/components/layout/SideNav/types.ts
```diff
@@ -1,6 +1,25 @@
...
     drawerWidth: number;
     setOpen: (open: boolean) => void;
 }
+
+export interface DrawerBoxProps {
+    isMediumScreen: boolean;
+    setOpen: (open: boolean) => void;
+}
+
+export interface IconAssignments {
```
NatSquared, line 15: This seems like a lot of trouble to go to... I'm thinking you should just extend the `interface Route` to include an `icon?: IconDefinition` which you can then render in your list mapping (or the broken link otherwise). Let me know if there'd be an issue with that

NatSquared: Also, aren't these other strings redundant if you have `[k: string]`?

met-web/src/components/layout/SideNav/SideNav.tsx
```diff
@@ -28,40 +40,142 @@ const DrawerBox = () => {
...
+        }
+        // Otherwise return the icon for the icon assignments key name that matches the route name.
+        return <FontAwesomeIcon icon={iconAssignments[route.name]} style={fontAwesomeStyles} />;
+    };
+
+    const CloseButton = () => {
```
NatSquared, line 120: Sonarcloud is suggesting (and I tend to agree) that this component definition should be extracted from the DrawerBox component and have its onClick behaviour passed in. Otherwise, just create it inline and use `<When>`

met-web/src/components/layout/SideNav/SideNav.tsx
```diff
@@ -28,40 +40,142 @@ const DrawerBox = () => {
...
             </List>
         </Box>
     );
 };
 
-const SideNav = ({ open, setOpen, isMediumScreen, drawerWidth = 280 }: SideNavProps) => {
+const SideNav = ({ open, setOpen, isMediumScreen, drawerWidth = 300 }: SideNavProps) => {
     if (!drawerWidth) return <></>;
     return (
         <ThemeProvider theme={DarkTheme}>
```
NatSquared, line 181: The DarkTheme provider around this component no longer really makes sense and should be removed, since the menu has gone from a dark background to a light one. This will probably save you some unnecessary style overrides deeper in the component.

NatSquared (changes requested):

met-web/src/components/layout/SideNav/types.ts
jareth-whitney: I changed this based on your second comment, I didn't realize that I could just define the structure and not the keys.

met-web/src/components/layout/SideNav/SideNav.tsx
jareth-whitney: Thanks, I removed it.

met-web/src/components/layout/SideNav/SideNav.tsx
jareth-whitney: No problem, I changed this so that the component is not within the DrawerBox.

NatSquared (approved):
Thank you for addressing my comments! If you could merge #2566 into this before it goes to `main` I'd appreciate it <3

met-web/src/components/layout/SideNav/SideNav.tsx
```diff
@@ -28,43 +75,118 @@ const DrawerBox = () => {
...
+        const possiblePageTitles: string[] = Object.keys(iconAssignments).filter((key) => key.includes(route.name));
+        if (0 === possiblePageTitles.length) {
+            return <FontAwesomeIcon icon={faLinkSlash} style={fontAwesomeStyles} />;
+        }
+        // Otherwise return the icon for the icon assignments key name that matches the route name.
+        return <FontAwesomeIcon icon={iconAssignments[route.name]} style={fontAwesomeStyles} />;
+    };
+
     return (
         <Box
```
Baelx, line 162: Nice making this menu a list in HTML! Since this is a [page region](https://www.w3.org/WAI/tutorials/page-structure/regions/), assistive technologies can benefit from having it marked as such. The below code will make the outer element a `<nav>` and give it a label.

https://www.w3.org/WAI/tutorials/page-structure/regions/#navigation

```suggestion
        <Box
            component="nav"
            aria-labelledby="Editor navigation"
```

met-web/src/components/layout/SideNav/UserGuideNav.tsx
```diff
@@ -63,9 +66,46 @@ const UserGuideNav = () => {
...
+                <FontAwesomeIcon
+                    icon={faBookOpen}
+                    style={{
+                        fontSize: '1.1rem',
+                        color: Palette.text.primary,
+                        paddingRight: '0.75rem',
+                        width: '1.1rem',
+                    }}
+                />
+                <MetHeader4
```
Baelx, line 91: Folks using assistive technologies will expect headers to denote a section to follow. If you just used an <h4> to make the text large and prominent, then you should use a link tag and style it to be large and prominent instead.

Also I'd strongly recommend we update the `openHelpPage` function to open the link in the same window. Using new windows/tabs can confuse some users and are discouraged unless preserving the current context of the page is absolutely necessary: https://www.w3.org/WAI/WCAG21/Techniques/general/G200

Baelx (changes requested):
I'm just seeing a few places where we can improve the accessibility experience. Can we discuss before you proceed?

- Make the parent HTML tag of the menu a `nav` and label it
- Change the Buttons in the menu to Links
- Don't use Heading tags in the nav menu unless they describe a section to follow. For large, prominent text, just use styling

met-web/src/components/layout/SideNav/SideNav.tsx
NatSquared: Aria-labelledby is for passing in an element ID; you want aria-label instead

Baelx: Thanks @NatSquared that's my bad. Also I think the label itself could probably be improved to something like "Admin navigation menu"

Baelx (approved):

PR #2566, opened 2024-07-25
[To Feature] DESENG-661 Patch: Optimizations and Accessibility

NatSquared:
Issue #: [🎟️ DESENG-661](https://citz-gdx.atlassian.net/browse/DESENG-661)

*Description of changes:*
- Removed "icon assignments" mapping and turned it into   a property of the Route item
- Renamed some variables to better indicate their purpose
- Added a more accessibly visible outline to route list items when navigating with the keyboard
- Made the entire route item surface clickable, expanding the <ListItemButton> component to take the full size of its parent. This should not affect the layout.

### DIFF
--- a/met-web/src/components/layout/SideNav/SideNav.tsx
+++ b/met-web/src/components/layout/SideNav/SideNav.tsx
@@ -13,25 +13,36 @@ import { faLinkSlash } from '@fortawesome/pro-regular-svg-icons/faLinkSlash';
 import { faCheck } from '@fortawesome/pro-solid-svg-icons/faCheck';
 import { faArrowLeft } from '@fortawesome/pro-solid-svg-icons/faArrowLeft';
 
+export const routeItemStyle = {
+    padding: 0,
+    backgroundColor: Palette.background.default,
+    '&:hover, &:focus': {
+        filter: 'brightness(96%)',
+    },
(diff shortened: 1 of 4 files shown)


met-web/src/components/layout/SideNav/SideNav.tsx
```diff
@@ -100,14 +91,25 @@ const DrawerBox = ({ isMediumScreen, setOpen }: DrawerBoxProps) => {
...
                     >
-                        {handleFontAwesomeIcon(route, itemType)}
+                        <FontAwesomeIcon
+                            style={{
+                                fontSize: '1.1rem',
+                                color: 'selected' === itemType ? Palette.primary.light : Palette.text.primary,
+                                paddingRight: '0.75rem',
+                                width: '1.1rem',
+                            }}
+                            icon={route.icon ?? faLinkSlash}
```
jareth-whitney, line 110: The double question mark paired with the integration of the icons with the route makes for a less verbose solution.

jareth-whitney (approved):
There are some good refactoring decisions here that will make the code easier to maintain. In the future, it would be great if we could communicate in advance regarding the reverse PRs, but I can't dispute that I'm benefiting from your work here.

PR #2568, opened 2024-07-30
[To Feature] DESENG-660 Part 2: Admin Header UI updates

NatSquared:
Issue #: [🎟️ DESENG-660](https://citz-gdx.atlassian.net/browse/DESENG-660)

*Description of changes:*
  - Redesigned the admin header to match the new design system
  - Added a new dropdown to the header to allow users to switch between tenants
  - Added a user menu, which for now only contains a logout button
  - Added a "drawer" view for the tenant switcher and user menu on mobile

This is part of a larger feature branch for DESENG-660

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the met-public license (Apache 2.0).

PR #2567, opened 2024-07-30
[To Feature] DESENG-660 Part 1: User/Tenant API changes

NatSquared:
Issue #: [🎟️ DESENG-660](https://citz-gdx.atlassian.net/browse/DESENG-660)

*Description of changes:*
  - Added a new API endpoint to fetch a list of a user's tenants, required for the new header

This is part of a larger feature branch for DESENG-660

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the met-public license (Apache 2.0).

PR #2569, opened 2024-07-31
[To Main] DESENG-660: Admin Header Redesign

NatSquared:
Issue #: [🎟️ DESENG-660](https://citz-gdx.atlassian.net/browse/DESENG-660)

✅ This PR consists of changes from:
#2567 (Approved)
#2568 (Approved)
[b6fa784](https://github.com/bcgov/met-public/pull/2569/commits/b6fa7846b1ecabf2ae7f06288ffa3b68d9469089) (Cosmetic changes only)

*Description of changes:*
- **Feature** New admin header
  - Added a new API endpoint to fetch a list of a user's tenants, required for the new header
  - Redesigned the admin header to match the new design system
  - Added a new dropdown to the header to allow users to switch between tenants
  - Added a user menu, which for now only contains a logout button
  - Added a "drawer" view for the tenant switcher and user menu on mobile


met-web/src/components/layout/Header/InternalHeader.tsx
```diff
@@ -1,104 +1,424 @@
...
+                        data-testid="sidenav-header"
+                        isMediumScreen={isMediumScreenOrLarger}
+                    />
+                </When>
+            </Box>
+        </TrapFocus>
+    );
+};
+
+const TenantSelector = ({
```
Baelx, line 207: Thanks for breaking these components up instead of having one huge one.

met-web/src/components/layout/Header/InternalHeader.tsx
```diff
@@ -1,104 +1,424 @@
...
+                    >
+                        {tenantList}
+                    </MenuList>
+                </Drawer>
+            </>
+        );
+    }
+
+    return (
+        <DropdownMenu
```
Baelx, line 309: I believe this would also be a navigation menu since the user would be navigated away from their current page if they selected an option (as opposed to filling out a dropdown in a form). If that's the case, then I'd recommend we make this a `nav` element and label - which will allow it to show up in screen reader rotors.

met-web/src/components/layout/Header/InternalHeader.tsx
```diff
@@ -1,104 +1,424 @@
...
+                        zIndex: (theme: Theme) => theme.zIndex.drawer + 3, // render under app bar but above side nav
+                        '& .MuiDrawer-paper': {
+                            padding: '1rem',
+                            top: '6.5rem',
+                            backgroundImage: 'none',
+                            borderBottomRightRadius: '16px',
+                        },
+                    }}
+                >
+                    <MenuList
```
Baelx, line 287: I'd recommend we set this to be a `nav` element if possible. I think any time you use a menu to navigate across the site, it should have a `nav` tag and be aria labelled, ideally.

Baelx (approved):
I'll set to approve since the sprint deadline is tomorrow but if possible, could I ask that we ensure any nav menu is given the `nav` tag and an aria label? I've commented in a couple places where that might need to be implemented. See here for WCAG's guidelines: https://www.w3.org/WAI/tutorials/menus/structure/#identify-menus

PR #2581, opened 2024-08-24
Feature/deseng668: Added authoring side nav and bottom nav, modified routing, added authoring context, added base page for banner

jareth-whitney:
Issue #: https://citz-gdx.atlassian.net/browse/DESENG-668

**Description of changes:**

- **Feature** New authoring content section
  - Implemented authoring side nav
  - Implemented authoring bottom nav
  - Modified routing
  - Implemented authoring section context
  - Still working on skeletons
  - Merged with main
  - Merged with Nat's feature branch
  - Made some post-merge fixes

N/A


met-web/src/components/engagement/admin/create/authoring/AuthoringSideNav.tsx
```diff
@@ -0,0 +1,281 @@
...
+    const currentBaseRoute = getRoutes(Number(engagementId), tenant)
+        .map((route) => route.base)
+        .filter((route) => location.pathname.includes(route))
+        .reduce((prev, curr) => (prev.length > curr.length ? prev : curr));
+
+    const allowedRoutes = getRoutes(Number(engagementId), tenant).filter((route) => {
+        return !route.authenticated || route.allowedRoles.some((role) => permissions.includes(role));
+    });
+
+    const renderListItem = (route: Route, itemType: string, key: number, engagementId: number) => {
```
NatSquared, line 55: I feel like this could be a JSX element rather than a render function that returns an element - would allow us to use real JSX syntax within the `allowedRoutes.map()` call. It should also be extracted outside of the main DrawerBox component to avoid [typescript:S6478](https://sonarcloud.io/organizations/bcgov-sonarcloud/rules?open=typescript%3AS6478&rule_key=typescript%3AS6478)

met-web/src/routes/AuthenticatedRoutes.tsx
```diff
@@ -88,33 +93,58 @@ const AuthenticatedRoutes = () => {
...
+                        }}
+                        element={<AdminEngagementView />}
+                    />
+                    <Route index element={<Navigate to="details/config" />} />
+                    <Route path="details">
+                        <Route index element={<Navigate to="config" />} />
+                        {/* Wraps the tabs with the engagement title and TabContext */}
+                        <Route element={<AdminEngagementView />}>
+                            <Route path="config" element={<ConfigTab />} />
+                            <Route path="authoring" element={<AuthoringTab />}></Route>
```
NatSquared, line 122: Add a splat route up here to provide the navigation tabs for lost users
```suggestion
                            <Route path="authoring" element={<AuthoringTab />}></Route>
                            <Route path="*" element={<NotFound />} />
```

met-web/src/routes/AuthenticatedRoutes.tsx
```diff
@@ -88,33 +93,58 @@ const AuthenticatedRoutes = () => {
...
+                        <Route path="*" element={<NotFound />} />
+                    </Route>
+                    <Route
+                        path="config/edit"
+                        element={<EngagementConfigurationWizard />}
+                        action={engagementUpdateAction}
+                        handle={{
+                            crumb: () => ({ name: 'Configure' }),
+                        }}
+                    />
```
NatSquared, line 144: Move the `config/edit` route under `/details` like we discussed
```suggestion
                        </Route>
                        <Route
                            path="config/edit"
                            element={<EngagementConfigurationWizard />}
                            action={engagementUpdateAction}
                            handle={{
                                crumb: () => ({ name: 'Configure' }),
                            }}
                        />
                    </Route>
```

met-web/src/routes/AuthenticatedRoutes.tsx
```diff
@@ -88,33 +93,58 @@ const AuthenticatedRoutes = () => {
...
+                    <Route index element={<Navigate to="details/config" />} />
+                    <Route path="details">
+                        <Route index element={<Navigate to="config" />} />
+                        {/* Wraps the tabs with the engagement title and TabContext */}
+                        <Route element={<AdminEngagementView />}>
+                            <Route path="config" element={<ConfigTab />} />
+                            <Route path="authoring" element={<AuthoringTab />}></Route>
+                        </Route>
+                        <Route path="authoring" element={<AuthGate allowedRoles={[USER_ROLES.EDIT_ENGAGEMENT]} />}>
```
NatSquared, line 124: For use in InternalHeader.tsx
```suggestion
                        <Route path="authoring" id="authoring" element={<AuthGate allowedRoles={[USER_ROLES.EDIT_ENGAGEMENT]} />}>
```

met-web/src/components/layout/Header/InternalHeader.tsx
```diff
@@ -23,15 +23,17 @@ import { Link } from 'components/common/Navigation';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faBars, faChevronDown, faClose, faSignOut } from '@fortawesome/pro-regular-svg-icons';
 import UserService from 'services/userService';
-import { Await, useAsyncValue, useRouteLoaderData } from 'react-router-dom';
+import { Await, useAsyncValue, useRouteLoaderData, useParams } from 'react-router-dom';
```
NatSquared, line 26:
```suggestion
import { Await, useAsyncValue, useRouteLoaderData, useParams, useMatches } from 'react-router-dom';
```

met-web/src/components/layout/Header/InternalHeader.tsx
```diff
@@ -23,15 +23,17 @@ import { Link } from 'components/common/Navigation';
...
-import { When, Unless } from 'react-if';
+import { When, Unless, If, Else, Then } from 'react-if';
 import { Button } from 'components/common/Input';
 import DropdownMenu, { dropdownMenuStyles } from 'components/common/Navigation/DropdownMenu';
 import { elevations } from 'components/common';
 import TrapFocus from '@mui/base/TrapFocus';
 import SideNav from '../SideNav/SideNav';
 import { USER_ROLES } from 'services/userService/constants';
+import AuthoringSideNav from '../../engagement/admin/create/authoring/AuthoringSideNav';
+import { getAuthoringRoutes } from '../../engagement/admin/create/authoring/AuthoringNavElements';
```
NatSquared, line 36:
```suggestion
```

met-web/src/components/layout/Header/InternalHeader.tsx
```diff
@@ -58,6 +60,19 @@ const InternalHeader = () => {
...
+    const pathname = window.location.href;
+    const { engagementId } = useParams() as { engagementId: string };
+    const currentAuthoringSlug = pathname.split('/').slice(-2).join('/');
+    const authoringRoutes = getAuthoringRoutes(Number(engagementId), tenant).map((route) => {
+        // skip the "Engagement Home" link
+        if ('Engagement Home' !== route.name) {
+            const pathArray = route.path.split('/');
+            return pathArray.slice(-2).join('/');
+        }
+    });
```
NatSquared, line 74:
```suggestion
    // Detect if the "authoring" route is a parent route of the current route
    // This is used to determine if the authoring sidenav should be displayed
    const matches = useMatches();
    const { engagementId } = useParams() as { engagementId: string };
    const isAuthoringRoute = matches.filter((match) => match.id === 'authoring').length > 0;
```

NatSquared (changes requested):
Hi Jareth - thanks for all your work on these components.
I've made some suggestions based on how I think you're trying to leverage these libraries; feel free to reach out if anything is unclear, or if I've missed something in my review.

met-web/src/styles/Theme.ts
```diff
@@ -356,13 +356,23 @@ export const DarkTheme = createTheme({
...
         MuiSelect: {
             defaultProps: {
-                MenuProps: {
-                    sx: {
-                        '&:hover': {
-                            backgroundColor: DarkPalette.hover.light,
-                        },
-                    },
-                },
+                // MenuProps: {
```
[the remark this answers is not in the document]
jareth-whitney: Removed. I can't account for this code as I didn't work on the DarkTheme component, but if anyone needs it they can grab it back from Github.

met-web/src/routes/AuthenticatedRoutes.tsx
jareth-whitney: Using my original approach to sidenav menu switching, as we discussed in Teams subsequent to this comment.

met-web/src/components/layout/Header/InternalHeader.tsx
jareth-whitney: See above comment regarding sidenav switching.

met-web/src/components/layout/Header/InternalHeader.tsx
jareth-whitney: See above comment regarding sidenav switching.

met-web/src/components/layout/Header/InternalHeader.tsx
jareth-whitney: See above comment regarding sidenav switching.

NatSquared (approved):

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+            <Grid sx={{ ...formItemContainerStyles, backgroundColor: colors.surface.blue[10] }} item>
+                <label htmlFor="section_heading">
+                    <MetHeader3 style={metHeader3Styles}>
+                        Section Heading <span style={{ fontWeight: 'normal' }}>(Required)</span>
+                    </MetHeader3>
+                    <EyebrowText style={eyebrowTextStyles}>
+                        Your section heading should be descriptive, short and succinct.
+                    </EyebrowText>
+                    <TextField
```
Baelx, line 108: Nat once mentioned that our TextField component attempts to auto generate a label. Could you let me know what the label for this text field will be ? Also, will screen reader users understand this field is required too?

If there's no label here then I'd recommend we make the h3 above into one and just style it to match the designs

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
```diff
@@ -0,0 +1,281 @@
...
+                        onChange={(value) => {
+                            setSectionHeading(value);
+                        }}
+                    />
+                </label>
+            </Grid>
+
+            <Grid sx={{ ...formItemContainerStyles, backgroundColor: colors.surface.gray[10] }} item>
+                <label htmlFor="body_copy">
+                    <MetHeader3 style={metHeader3Styles}>
```
Baelx, line 124: Even though this resembles a heading in the design, functionally it's more of a label. Same question as above, do screen readers get enough information about this field in order to fill it in?

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
```diff
@@ -0,0 +1,281 @@
...
+                        Body Copy
+                        <span style={{ fontWeight: 'normal' }}> (Required)</span>
+                    </MetHeader3>
+                    <EyebrowText style={eyebrowTextStyles}>
```
Baelx, line 128: We're giving sighted users all this nice information about what goes in this field. Is it easy for screen reader users to perceive this as well?

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
```diff
@@ -0,0 +1,281 @@
...
+                    />
+                </label>
+            </Grid>
+
+            <Grid sx={{ ...formItemContainerStyles, backgroundColor: colors.surface.blue[10] }} item>
+                <label htmlFor="survey_button_text">
+                    <MetHeader3 style={metHeader3Styles}>
+                        Survey Call to Action Button <span style={{ fontWeight: 'normal' }}>(Optional)</span>
+                    </MetHeader3>
+                    <EyebrowText style={eyebrowTextStyles}>
```
Baelx, line 154: My understanding of an "eyebrow" is that it's an additional piece of info that goes _above_ the "eye" - in this case the heading or label. I kind of feel this should be a label itself though

met-web/src/components/engagement/admin/create/authoring/AuthoringTemplate.tsx
```diff
@@ -0,0 +1,192 @@
...
+                            {/* <StatusLabel text={'Insert Section Status Text Here'} completed={'Insert Completed Boolean Here'} /> */}
+                        </div>
+                    )}
+                </Await>
+            </Suspense>
+            <h1 style={{ marginTop: '0.5rem', paddingBottom: '1rem' }}>{pageTitle}</h1>
+            <SystemMessage status="warning">
+                Under construction - the settings in this section have no effect.
+            </SystemMessage>
+            {'details' === slug && (
```
Baelx, line 111: Just curious why this markup lives here instead of in AuthoringDetails.jsx. Was it easier to style or something?

met-web/src/components/engagement/admin/create/authoring/AuthoringTemplate.tsx
```diff
@@ -0,0 +1,192 @@
...
+                        )}
+                    </Await>
+                </Suspense>
+
+                <UnsavedWorkConfirmation blockNavigationWhen={isDirty && !isSubmitting} />
+                <Suspense>
+                    <Await resolve={languages}>
+                        {(languages: Language[]) => (
+                            <AuthoringBottomNav
+                                isDirty={isDirty}
```
Baelx, line 176: Nice clean code using form context to pass isDirty into the bottom nav ^_^

met-web/src/components/engagement/admin/create/authoring/engagementAuthoringUpdateAction.tsx
```diff
@@ -24,20 +24,22 @@ export const engagementAuthoringUpdateAction: ActionFunction = async ({ request
...
-        console.error('Error updating engagement', e);
-        errors.push(e);
+    if ((formData.get('title') || formData.get('icon_name')) && '0' !== formData.get('content_id')) {
+        try {
+            await patchEngagementContent(engagement.id, Number(formData.get('content_id')) as unknown as number, {
+                title: formData.get('title') as string,
+                icon_name: formData.get('icon_name') as string,
+            });
+        } catch (e) {
+            console.error('Error updating engagement', e);
```
Baelx, line 34: Thanks for the descriptive error in the console call!

Baelx (changes requested):
Great work on this Jareth! You've done tons of work here. 

I just want to ensure screen reader users are getting the same experience as others. Can you respond to some of my comments about the use of elements and how they help screen readers? Also, regarding the Preview button, I think we should use an icon button instead there.

met-web/src/components/engagement/admin/create/authoring/AuthoringTemplate.tsx
jareth-whitney: It's just about page positioning. For the details page specifically, there is a "content configuration" that appears above the current language h2. We could change that and have the content local to the Details component.

met-web/src/components/engagement/admin/create/authoring/AuthoringTemplate.tsx
jareth-whitney: thanks!

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
jareth-whitney: I imported this eyebrow text component and used it because it resembled the type in the design, even though the name might not be 100% accurate. I could import it AS something else if you think it's a bit confusing.

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
jareth-whitney: That's a good question! It's harder to test screen reader compatibility through Firefox without ChromeVox, but I found a plugin called Pericles and tested it with the page. The text was read by the screenreader, so we should be ok.

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
jareth-whitney: I just looked and I'm not seeing an auto-generated label for the TextField component in the browser. Perhaps I need to use a specific prop. If it's ok with you, I think this is something that could be addressed in the individual authoring page tickets.

met-web/src/components/engagement/admin/create/authoring/AuthoringContext.tsx
```diff
@@ -26,55 +27,67 @@ export const AuthoringContext = () => {
...
             }),
             {
                 method: 'post',
-                action: `/engagements/${data.id}/authoring/${slug}`,
+                action: `/engagements/${data.id}/details/authoring/${slug}`,
             },
         );
     };
 
+    const getSendReportValue = (valueToInterpret: boolean) => {
```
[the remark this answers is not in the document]
jareth-whitney: So the data in createSearchParams all needs to be converted to strings. In this particular case, I just needed to convert a boolean value to a string equivalent as 'true' or 'false'. It's quite possible that there is another utility function that does this, I just don't know about it.

met-web/src/components/engagement/admin/create/authoring/AuthoringBottomNav.tsx
```diff
@@ -141,7 +117,14 @@ const AuthoringBottomNav = (props: AuthoringBottomNavProps) => {
                                 marginLeft: 'auto',
                             }}
                         >
-                            <img style={{ paddingRight: '0.3rem' }} src={pagePreview} alt="page preview icon" />
+                            <img
```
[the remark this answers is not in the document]
jareth-whitney: I'm not sure how to make it into an icon since this is a custom graphic that Jessica created for the button. I exported it from Figma and included it in our assets. If we have a process for converting images to icons then let me know and I will do it.

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
jareth-whitney: It would be read as a heading 3 at the moment, but that could definitely be changed if you think that's confusing to screen readers. Firefox doesn't have the best support for screen readers but I am testing the pages briefly with Pericles.

met-web/src/components/engagement/admin/create/authoring/AuthoringContext.tsx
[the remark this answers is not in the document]
jareth-whitney: Booleans can't be used with truthy checks because they can return false if their value is false, even though I want that conditional to pass on a false value. So I need to check if the value is undefined separately, then give a value that is either 'true' or 'false'.

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
Baelx: I'd suggest using Mac OS' built-in Voiceover as it's one of the most used screen readers. It's really important that we test our work with a screen reader to ensure those users are having a good experience too

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
Baelx: Thanks for testing that! It's important not only that a screen reader can read it but also that it's clear that this information is associated with a particular form field. When you tabbed your focus to this field, did you hear all this text read out?

met-web/src/components/engagement/admin/create/authoring/AuthoringTemplate.tsx
Baelx: Just looked back at the designs and that makes sense!  Thanks for explaining

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
Baelx: It's not necessarily confusing but it's important that we don't just code for looks but that our HTML is semantic and meaningful. In this case, I almost feel like both the metheader3 and eyebrowtext elements should be labels that describe the form field.

met-web/src/components/engagement/admin/create/authoring/AuthoringBottomNav.tsx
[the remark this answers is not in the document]
jareth-whitney: I looked it up and apparently I can add alt="" and aria-hidden="true" in order to keep it invisible to a reader.

met-web/src/components/engagement/admin/create/authoring/AuthoringContext.tsx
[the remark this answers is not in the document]
Baelx: Yeah I was just curious how this operation for creating an HTTP request gets handled elsewhere and if we could use that. No worries if you want to just stick with this function you've created

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
Baelx: I appreciate that you've already done a lot of work for the authoring flow but I think it'd be best to handle these now. We already have to update the other tickets to describe what's been done in this one and I'm worried about adding another complicating factor like "rework the HTML to better serve screen readers." If we're publishing the HTML then we should do it right from the get-go.

That said, I don't think it'd be too much work to change each of those elements to a label, updating the styling if needed, then associate them with the input fields. It should be the same fix for each of the pages. This would be one approach.

What's important is just that screen readers read out all this information when a user tabs onto a field - however you want to approach that

met-web/src/components/engagement/admin/create/authoring/AuthoringBottomNav.tsx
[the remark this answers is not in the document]
Baelx: I'm alright with this as a quick fix this time. Could you please just test it once you implement?

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
jareth-whitney: I have removed all the MetHeader3 components and replaced them with MetLabel components that are imported as MetBigLabel. This means that they show up as <p> in the DOM instead of <h3>. Anything that is a real h3 heading is still an h3. I renamed EyebrowText to FormDescriptionText so that it's clear what its use is. This text is all already contained within a label element so it's part of the label, as is the form component itself.

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
jareth-whitney: The tab just cycles through links, I believe in ChromeVox to cycle through individual items it's something like CTRL+ALT+CMD+DOWN ARROW (or up arrow). The screen reader reads the entire page including this text otherwise.

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
jareth-whitney: Thanks, I tested with this too.

met-web/src/components/engagement/admin/create/authoring/AuthoringFeedback.tsx
jareth-whitney: I made some other changes that will definitely help the way screen readers read the page. I've now tested it with Pericles and the mac OS voice service. The h3s that were form labels have been converted to form labels. I also removed the alt tag on the image in the "Preview" button and added an aria-invisible tag to it.

Baelx (approved):
Thanks for all your work on this! Approved

PR #2595, opened 2024-09-26
Feature/deseng692: Reverted to previous title logic for video widget

jareth-whitney:
Issue #: https://citz-gdx.atlassian.net/browse/DESENG-692

**Description of changes:**
- Removed unneeded columns from db (title from widget_video table and associated translation column)
- Updated all other met_api and met_web logic to accomodate this
- Title is now pulled from widget table, not widget_video table

N/A


Baelx (approved):
Thanks for saving us some code and using existing fields where we can!

PR #2606, opened 2025-04-23
DESENG-775 Update dagster and python container version

Baelx:
Issue #: https://citz-gdx.atlassian.net/browse/DESENG-775

**Description of changes:**
- Updated the versions for dagster, it's associated libraries, and the python container

**User Guide update ticket (if applicable):** N/A


Baelx:
I've done this from main instead of from the `omega` project. Going to re-do

PR #2819, opened 2026-04-09
Hotfix: add pytz to api "prod" requirements

NatSquared:
Issue #: N/A

**Description of changes:**
Hotfix an issue in `cron` where `pytz` was not being installed, causing crashes like the below.
This is because in `api/setup.py`, only `prod.txt` is installed as requirements for the `api` package.
Also upgrade `pytz` to 2026.1.post1 globally for consistency (BC has a new time zone!)

image


jareth-whitney (approved):
Looks like you just needed to update a dependency version, good to go.

PR #2835, opened 2026-04-20
Fix Upcoming and Closed states in hero banner

NatSquared:
Issue #: [🎟️ DEP-260](https://citz-gdx.atlassian.net/browse/DEP-260)

**Description of changes:**
- **Bugfix** Ensure Upcoming and Closed state texts appear in the hero banner (and nowhere else)
  - Fixed a bug where the Upcoming and Closed state texts could appear in the Provide Feedback section of the engagement details page, which was not the intended behavior. These status texts are now only displayed in the hero banner of the engagement page when the engagement is in the corresponding status.

**Common component changes:**

### DIFF
--- a/web/src/components/engagement/public/view/EngagementHero.tsx
+++ b/web/src/components/engagement/public/view/EngagementHero.tsx
@@ -6,7 +6,11 @@ import { getEditorStateFromRaw } from 'components/common/RichTextEditor/utils';
 import { Engagement } from 'models/engagement';
 import { Box, Grid2 as Grid, Skeleton } from '@mui/material';
 import { colors } from 'components/common';
-import { EngagementStatusChip, getSubmissionStatusFromPreviewState } from 'components/common/Indicators';
+import {
+    EngagementStatusChip,
+    getStatusFromStatusId,
+    getSubmissionStatusFromPreviewState,
+} from 'components/common/Indicators';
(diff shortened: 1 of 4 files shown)


web/src/components/engagement/public/view/EngagementHero.tsx
```diff
@@ -185,46 +197,45 @@ export const EngagementHero = () => {
...
-                                                size="large"
-                                                icon={<FontAwesomeIcon fontSize={24} icon={faChevronRight} />}
-                                                iconPosition="right"
-                                                sx={{ borderRadius: '8px' }}
-                                            >
-                                                {buttonContent.label}
-                                            </Button>
-                                        )}
-                                    </PreviewRender>
+                                    {(heroButtonContent || previewHeroButtonFallback) && (
```
jareth-whitney, line 218: I'm guessing the fallback value is asynchronous as well, otherwise we could eliminate this check.

jareth-whitney (approved):
Looks good, should clean up the look of the engagement page under those specific conditions.

web/src/components/engagement/public/view/EngagementHero.tsx
NatSquared: It's not that it's asynchronous, it just could be null (such as when we are not previewing the page). In that case, we do not render the preview

PR #2842, opened 2026-04-30
DEP-261: update object storage service to normalize URLs to keys

NatSquared:
Issue #: [🎟️ DEP-261](https://citz-gdx.atlassian.net/browse/DEP-261)

**Description of changes:**
- **Feature** Updated all S3 uploaded objects to dynamically generate document URLs from the bucket and object key, rather than storing the full URL in the database. 
  - Updated the API to generate S3 object URLs on-the-fly using the bucket name and object key, rather than storing the full URL in the database. This allows for more flexibility in changing storage configurations in the future without needing to update existing database records.
  - Updated relevant API endpoints to reflect this change, including those for retrieving engagement details, survey details, and any other endpoints that return data with S3 object references.
  - Added basic unit tests for S3 service and URL generation test to ensure that URLs are generated correctly.

**Common component changes:**


api/tests/unit/services/test_object_storage_service.py
```diff
@@ -0,0 +1,158 @@
...
+@patch('api.services.object_storage_service.AWSRequestsAuth', side_effect=_FakeAWSAuth)
+@patch('api.services.object_storage_service.Config')
+def test_get_url_returns_empty_when_filename_missing(mock_config, _mock_auth):
+    """Assert URL construction fails safely when filename/key is missing."""
+    mock_config.return_value.S3_CONFIG = TEST_S3_CONFIG
+
+    service = ObjectStorageService()
+
+    assert service.get_url('') == ''
+    assert service.get_url(None) == ''
```
jareth-whitney: It looks like SonarCloud doesn't like the type that you're passing to get_url here.

web/src/components/tenantManagement/TenantForm.tsx
```diff
@@ -86,7 +86,7 @@ export const TenantForm = ({
         }
         try {
             const savedDocumentDetails = await saveObject(bannerImage, { filename: bannerImage.name });
-            return savedDocumentDetails?.filepath || '';
+            return savedDocumentDetails?.uniquefilename || '';
```
jareth-whitney, line 89: Nice, I prefer this

jareth-whitney (approved):
Well that's a fancy way to make links - looks good! Just review what you're passing into get_url() on line 96 of test_object_storage_service.py.

PR #2843, opened 2026-05-05
DEP-262: add SUPER_ADMIN role and associated functionality to db; remove staff_user tenant_id column

NatSquared:
Issue #: [🎟️ DEP-262](https://citz-gdx.atlassian.net/browse/DEP-262)

**Description of changes:**

- **Feature** Sync SUPER_ADMIN tenant membership and protect super admins in user management
  - Added a migration to create the `SUPER_ADMIN` group, the `super_admin` role, and their mapping so super admins can be represented in tenant memberships.
  - Updated the API to add or remove `SUPER_ADMIN` group membership based on the user's Keycloak roles during login/request processing, including unit tests for assignment and removal.
  - Updated the user management UI to recognize Super Admin as a composite role and prevent administrators and super admins from being deactivated, while showing the correct permission messaging.
  - Removed the tenant_id column from staff_user table, as tenant membership is determined by group memberships rather than a direct column on the user model, making the column redundant.

**Common component changes:**

### DIFF
--- a/CHANGELOG.MD
+++ b/CHANGELOG.MD
@@ -1,3 +1,11 @@
+## May 4, 2026
+
+- **Feature** Sync SUPER_ADMIN tenant membership and protect super admins in user management [🎟️ DEP-262](https://citz-gdx.atlassian.net/browse/DEP-262)
+  - Added a migration to create the `SUPER_ADMIN` group, the `super_admin` role, and their mapping so super admins can be represented in tenant memberships.
+  - Updated the API to add or remove `SUPER_ADMIN` group membership based on the user's Keycloak roles during login/request processing, including unit tests for assignment and removal.
+  - Updated the user management UI to recognize Super Admin as a composite role and prevent administrators and super admins from being deactivated, while showing the correct permission messaging.
+  - Removed the tenant_id column from staff_user table, as tenant membership is determined by group memberships rather than a direct column on the user model, making the column redundant.
+
 ## April 30, 2026
 
 - **Feature** Updated all S3 uploaded objects to dynamically generate document URLs from the bucket and object key, rather than storing the full URL in the database. [🎟️ DEP-261](https://citz-gdx.atlassian.net/browse/DEP-261)
--- a/api/migrations/versions/9d2d16f5f3aa_add_super_admin_role_group_mapping.py
+++ b/api/migrations/versions/9d2d16f5f3aa_add_super_admin_role_group_mapping.py
@@ -1,0 +1,80 @@
+"""Add super admin role/group mapping.
+
+Revision ID: 9d2d16f5f3aa
+Revises: 8e50e47b7d32
+Create Date: 2026-04-30 12:00:00.000000
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '9d2d16f5f3aa'
+down_revision = '8e50e47b7d32'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        """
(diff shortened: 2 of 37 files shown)


api/migrations/versions/9d2d16f5f3aa_add_super_admin_role_group_mapping.py
```diff
@@ -0,0 +1,80 @@
...
+              SELECT 1
+              FROM group_role_mapping m
+              WHERE m.group_id = g.id
+                AND m.role_id = r.id
+          )
+        """
+    )
+
+
+def downgrade():
```
jareth-whitney, line 57: So remove all user_group_memberships where the role ID is equal to super_admin. Yes that makes sense.

api/migrations/versions/a6f15ee2c7b4_remove_staff_user_tenant_id.py
```diff
@@ -0,0 +1,48 @@
...
+revision = 'a6f15ee2c7b4'
+down_revision = '9d2d16f5f3aa'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    _drop_matching_foreign_keys('staff_users', 'tenant_id', 'tenant')
+    with op.batch_alter_table('staff_users', schema=None) as batch_op:
+        batch_op.drop_column('tenant_id')
```
jareth-whitney, line 22: Are we removing tenant ID from the staff_users table because it's a redundant link? I.e. you can extract the users from the tenant, therefor the tenant doesn't need to be nested in the user? Otherwise it seems like it could be useful to have available.

api/tests/conftest.py
```diff
@@ -220,7 +220,7 @@ def setup_team_member_and_claims(jwt):
     """Set up a user with the team member role."""
     staff_info = dict(TestUserInfo.user_staff_1)
     user = factory_staff_user_model(user_info=staff_info)
-    factory_user_group_membership_model(str(user.external_id), user.tenant_id, CompositeRoleId.TEAM_MEMBER.value)
+    factory_user_group_membership_model(str(user.external_id), group_id=CompositeRoleId.TEAM_MEMBER.value)
     claims = copy.deepcopy(TestJwtClaims.team_member_role.value)
     claims['sub'] = str(user.external_id)
```
jareth-whitney, line 226: Thank you for updating the unit tests thoroughly.

CHANGELOG.MD
```diff
@@ -1,3 +1,11 @@
+## May 4, 2026
+
+- **Feature** Sync SUPER_ADMIN tenant membership and protect super admins in user management [🎟️ DEP-262](https://citz-gdx.atlassian.net/browse/DEP-262)
+  - Added a migration to create the `SUPER_ADMIN` group, the `super_admin` role, and their mapping so super admins can be represented in tenant memberships.
+  - Updated the API to add or remove `SUPER_ADMIN` group membership based on the user's Keycloak roles during login/request processing, including unit tests for assignment and removal.
```
jareth-whitney, line 5: I think this should be safe if we're qualifying the information that we're receiving from Keycloak. I just want to make sure that there's no way a bad actor could spoof these credentials in order to gain super_admin access. If it's just textbook OAuth/SAML then please ignore.

jareth-whitney (approved):
This is great, I think it will help with some of the login issues I've been experiencing. No dealbreakers here, you're good to merge. 

I believe, from reading your PR, that we will always have a tenant_id for a user based on data from another table or app state, but please verify that's the case. Also, can you imagine a scenario where a bad actor could spoof their Keycloak role, triggering super_admin privilege? I'm guessing not, but it's worth thinking about for a moment.

PR #2891, opened 2026-07-07
feature/dep315: Modified app layout, added new landing components, refactored landing components, other misc fixes.

jareth-whitney:
…

Issue #: https://citz-gdx.atlassian.net/browse/DEP-315

**Description of changes:**

- **Feature** Integrated new landing hero section and introduction
    - Refactored existing landing code for structural simplicity
    - Created a reusable container/section block that can be reused anywhere
    - Added params to URL for future filter updating (can be decoupled easily if not desired)
    - Implemented UX designs (mobile and desktop)
    - Removed app's square box that appears at bigger viewports
    - Ensured that other layouts were not negatively affected by change (admin, admin landing preview, admin engagement preview)
        - Added constrained option for internal header component as an optional boolean prop
- Fixed: Engagement preview was looking for route loader address that didn't exist
- Fixed: Slug was not being populated with SuggestedEngagementAttachment, so Suggested Engagement links were not working
- Fixed: Small typo in security response headers of API
- Fixed: Console error for authoring section details tabs, selection/list mismatch
- Enhanced: Notification banner now supports multiple lines for mobile (or long messages, if needed in the future)

- - [X] No, a user guide update ticket is not required.

**Common component changes:**

- - [X] No, there are no changes to the common components.


web/src/components/engagement/preview/EngagementPreview.tsx
```diff
@@ -186,7 +186,8 @@ const MeasurementBar: React.FC = () => {
  * modifying the actual engagement.
  */
 export const EngagementPreview: React.FC = () => {
-    const loaderData = useRouteLoaderData('public-single-engagement') as EngagementLoaderPublicData;
+    const loaderData = useRouteLoaderData('engagement-preview') as EngagementLoaderPublicData;
```
NatSquared: thanks for catching this incorrect loader call - if you don't want to add an ID in AuthenticatedRoutes.tsx every time, you can also switch this to a `useLoaderData()` and it'll fetch data from the nearest ancestor route with a loader

web/src/DocumentTitle.tsx
```diff
@@ -39,7 +39,6 @@ const DocumentTitle = () => {
         const resolveTitle = async () => {
             const crumbs = await getResolvedCrumbs(matches);
             if (cancelled) return;
-            console.log('Resolved crumbs for title:', crumbs);
```
NatSquared, line 42: thanks for the catch :P

web/tests/unit/components/landingPage/LandingPage.test.tsx
```diff
@@ -37,7 +51,6 @@ jest.mock('constants/engagementStatus', () => ({
         Upcoming: 5,
         Open: 6,
         Unpublished: 7,
-        // Allow backwards lookup like the enum we're mocking
         1: 'Draft',
         2: 'Published',
```
NatSquared, line 59: Thanks for refactoring these tests! Any reason why the existing comments were removed throughout?

web/src/components/layout/Header/EnvironmentBanner.tsx
```diff
@@ -17,9 +18,10 @@ const EnvironmentBanner = () => {
             container
             direction="row"
             gap={2}
-            padding={{ xs: '0 1em 0', md: '0 5vw 0', lg: '0 10em 0' }}
+            padding={`1rem ${Layout.padding.default}`}
```
NatSquared: Is the environment banner still compliant with the Figma design after adding 1rem of vertical padding? That should probably still be `0`

web/src/components/landing/LandingSection.tsx
```diff
@@ -0,0 +1,33 @@
...
+const LandingSection = ({ children, image, colour, outerStyles = {}, innerStyles = {} }: LandingSectionProps) => {
+    const outerGridStyles = {
+        backgroundImage: image ? `url(${image})` : undefined,
+        backgroundSize: 'cover',
+        backgroundPosition: 'center',
+        backgroundColor: image ? 'transparent' : colour || colors.surface.white,
+        width: '100%',
+        padding: Layout.padding.default, // Outer padding
+        ...outerStyles,
+    };
```
NatSquared: Inline styling might be easier to read here, since we are not doing anything besides styling in this component. The `width` and `padding` properties can be passed directly to the component instead of as keys of the `sx` object. Also, you can avoid spreading outerStyles by passing an array into sx (see below)

web/src/components/landing/LandingSection.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const innerGridStyles = {
+        margin: '0 auto', // Centre content
+        width: '100%', // Becomes 100% width under max width value
+        maxWidth: Layout.width.default, // Maximum width of content container
+        ...innerStyles,
+    };
+
+    return (
+        <Grid container direction="column" sx={outerGridStyles}>
+            <Grid container direction="column" sx={innerGridStyles}>
```
NatSquared:
```suggestion
    };

    return (
        <Grid container direction="column" sx={[outerGridStyles, outerStyles]}>
            <Grid container direction="column" sx={[innerGridStyles, innerStyles]}>
```

web/src/components/landing/LandingSection.tsx
```diff
@@ -0,0 +1,33 @@
...
+        backgroundColor: image ? 'transparent' : colour || colors.surface.white,
+        width: '100%',
+        padding: Layout.padding.default, // Outer padding
+        ...outerStyles,
+    };
+
+    const innerGridStyles = {
+        margin: '0 auto', // Centre content
+        width: '100%', // Becomes 100% width under max width value
+        maxWidth: Layout.width.default, // Maximum width of content container
```
NatSquared: All of these properties can be set on the component itself; using `sx` is not needed here.

web/src/components/landing/index.tsx
```diff
@@ -1,12 +1,105 @@
...
+        filtersOpen,
+        setFiltersOpen,
+    };
+
+    const engSearchProps = {
+        ...filterProps,
+        engagements,
+        loadingEngagements,
+        totalEngagements,
+        searchFilters,
```
NatSquared: `searchFilters` and `setSearchFilters` are added both in `filterProps` and again in `engSearchProps`

web/src/components/landing/index.tsx
```diff
@@ -1,12 +1,105 @@
...
+            include_banner_url: defaultSearchParams.include_banner_url,
+            size: defaultSearchParams.size,
+        });
+        setSearchParams(newSearchParams);
+    };
+
+    const clearFilters = () => {
+        setMetadataFilters([]);
+        setSearchFilters(defaultSearchFilters);
+    };
```
NatSquared, line 33: We probably don't want to clear the list of *available* filters; just the ones the user is currently filtering by.
```suggestion
    const clearFilters = () => {
        setSearchFilters(defaultSearchFilters);
    };
```

web/src/components/landing/index.tsx
```diff
@@ -1,12 +1,105 @@
...
+        });
+    }, [loaderData.engagements]);
+
+    useEffect(() => {
+        loaderData.filters.then((filters: MetadataFilter[]) => {
+            if (filters && filters.length > 0) {
+                setMetadataFilters(filters);
+            }
+        });
+    }, [loaderData.filters]);
```
NatSquared: I notice you're using a useState here instead of awaiting the data in the consumer component; were you not able to get the filters working with the loader pattern?

web/src/components/landing/index.tsx
NatSquared: We could probably still make use of a handful of context values instead of prop-drilling here (as long as we're not using the useState loading pattern), but it's up to you

NatSquared (approved):
Looks good; thanks for refreshing the landing page. Just a few things to review

web/src/components/engagement/preview/EngagementPreview.tsx
jareth-whitney: Yes, that should work fine here.

web/src/components/landing/index.tsx
jareth-whitney: I'll see if I can pass down the promise directly.

jareth-whitney: Also I refactored the filters and got them working

web/src/components/landing/index.tsx
jareth-whitney: Right you are, I renamed the variables to make it more clear that we are managing both ALL of the filters and which ones are selected.

web/src/components/landing/index.tsx
jareth-whitney: I overhauled the system so that we are not following state in two places anymore, and we are using one context to pass down all of the values, no prop drilling.

web/src/components/landing/LandingSection.tsx
jareth-whitney: I did try using the component properties, but I found that they are not always equivalent to the CSS, which is why I used sx props here instead.

web/src/components/landing/LandingSection.tsx
jareth-whitney: I can switch it to inline

web/src/components/landing/LandingSection.tsx
jareth-whitney: I have the styles inline now, so everything will be in the return.

web/src/components/layout/Header/EnvironmentBanner.tsx
jareth-whitney: It was pretty close, about 11px of space vs 16, so I changed it to 11px to be consistent. The thing is that before it had a fixed height of 50 px, so the padding was more... implied than literal. The text was centred in a heighted div. Once you switch to multiline you have to account for that spacing, because it's no longer managed by height. Anyway, it's now 11px instead of 1rem.

web/tests/unit/components/landingPage/LandingPage.test.tsx
jareth-whitney: I refactored the tests a bit but I'll add some comments back when I have a chance. For now it should pass with no issues.

web/src/components/landing/index.tsx
jareth-whitney: It will be irrelevant now because of the context and searchFilters are no longer tracked.

PR #2911, opened 2026-07-28
DEP-276: Formio styling and functionality adjustments

NatSquared:
Issue #: [🎟️ DEP-276](https://citz-gdx.atlassian.net/browse/DEP-276)

**Description of changes:**
- **Feature** Re-style Formio components; ensure BC Gov components work properly
  - Ensured all Formio imports come from the same package, which means all components are brought into the same runtime when using Formio.use(). This fixes an issue where some components were unable to render certain other components as children.
  - Updated Formio component styles to match BC Gov design system, including button sizes, focus states, and other visual elements. This ensures a consistent look and feel across the application. Also imported some style overrides from the CHEFS frontend.
  - Created custom overrides to ensure focus states and buttons feel at home within DEP.
  - Added focusability for buttons that lacked tabindex and aria roles, improving accessibility for keyboard and screen reader users.
  - Removed kludged Bootstrap styles and imported Bootstrap SCSS directly into the Formio SCSS file, allowing for better control over styling and ensuring Bootstrap does not overwrite outer components.
  - Removed buggy Typescript declarations for Formio.Utils; import directly from Formio instead
  - Moved admin survey preview page; adjusted link name

**Common component changes:**


web/src/components/Form/FormBuilder.tsx
```diff
@@ -1,15 +1,178 @@
...
+
+            if (event.key === 'Enter' || event.key === ' ') {
+                event.preventDefault();
+                pendingFocusIndex = currentIndex;
+                currentTab.click();
+                return;
+            }
+
+            const lastIndex = tabs.length - 1;
+            const nextIndex = (() => {
```
jareth-whitney, line 49: Some nice touches with key capture here

web/src/components/Form/FormBuilder.tsx
```diff
@@ -1,15 +1,178 @@
...
+            nextTab.click();
+        };
+
+        const handleAddPageKeydown = (event: KeyboardEvent) => {
+            if (event.key !== 'Enter' && event.key !== ' ') {
+                return;
+            }
+
+            event.preventDefault();
+            (event.currentTarget as HTMLElement).click();
```
jareth-whitney, line 89: Wondering if there needs to be any consideration for detached nodes or another scenario where event.currentTarget isn't defined.

web/src/components/Form/FormBuilder.tsx
```diff
@@ -1,15 +1,178 @@
...
+                control.setAttribute('tabindex', '0');
+                control.setAttribute('role', 'button');
+                control.setAttribute('aria-label', 'Add page');
+                control.onkeydown = handleAddPageKeydown;
+            });
+
+            if (pendingFocusIndex !== null) {
+                const tabToFocus = wizardTabs[pendingFocusIndex] ?? activeTab;
+                if (tabToFocus) {
+                    tabToFocus.focus();
```
jareth-whitney, line 141: optionally tabToFocus?.focus()

web/src/components/Form/FormBuilder.tsx
```diff
@@ -1,15 +1,178 @@
...
+        observer.observe(formioRoot, {
+            childList: true,
+            subtree: true,
+            attributes: true,
+            attributeFilter: ['class'],
+        });
+
+        return () => {
+            observer.disconnect();
+            getWizardPageTabs().forEach((tab) => {
```
jareth-whitney, line 181: Any chance this will come back blank? It looks like that function relies on a query selector to retrieve values. If they're always guaranteed because of your process then no worries.

web/src/components/Form/formio.scss
```diff
@@ -1,377 +1,277 @@
...
+    @import "~bootstrap/scss/reboot", "~bootstrap/scss/type", "~bootstrap/scss/containers", "~bootstrap/scss/grid", "~bootstrap/scss/tables", "~bootstrap/scss/forms", "~bootstrap/scss/buttons", "~bootstrap/scss/transitions", "~bootstrap/scss/dropdown", "~bootstrap/scss/button-group", "~bootstrap/scss/nav", "~bootstrap/scss/navbar", "~bootstrap/scss/card", "~bootstrap/scss/accordion", "~bootstrap/scss/breadcrumb", "~bootstrap/scss/pagination", "~bootstrap/scss/badge", "~bootstrap/scss/alert", "~bootstrap/scss/progress", "~bootstrap/scss/list-group", "~bootstrap/scss/close", "~bootstrap/scss/toasts", "~bootstrap/scss/modal", "~bootstrap/scss/tooltip", "~bootstrap/scss/popover", "~bootstrap/scss/carousel", "~bootstrap/scss/spinners", "~bootstrap/scss/offcanvas", "~bootstrap/scss/placeholders", "~bootstrap/scss/helpers", "~bootstrap/scss/utilities/api"; // NOSONAR
+    @import '~met-formio/dist/met-formio-components'; // NOSONAR
+    @import '@bcgov/formio/dist/bcgov-formio-components'; //NOSONAR
+}
+
+// Apply Formo and CHEFs styles after Bootstrap styles so they can override them as needed
+@import '~formiojs/dist/formio.full.min'; //NOSONAR
+@import 'CHEFS-style.scss'; //NOSONAR
+
```
jareth-whitney, line 21: It looks like a ton of additional styles had to be loaded to make this work properly, kudos for getting the cascading and dependencies down.

web/src/components/Form/formio.scss
```diff
@@ -1,377 +1,277 @@
...
+// Styling for multi-page form page selectors + drag & drop functionality 
+ol.breadcrumb.wizard-pages li,
+li.gu-mirror {
+    & .badge {
+        padding: 0.5rem;
```
jareth-whitney, line 26: SCSS looks so weird after using inline sx/styling objects in React

web/src/components/Form/formio.scss
```diff
@@ -1,377 +1,277 @@
...
+    overflow: visible; // allow acessibility borders to show outside of the component bounds
+    .choices__inner,
+    .choices__list.choices__list--dropdown {
+        &:has(:focus) {
+            border: 1px solid #7A7876;
+            outline: 4px solid #2E5DD7;
+        }
+        border-radius: var(--bs-border-radius);
+    }
+    &[data-type*="select-one"] .choices__button:focus {
```
jareth-whitney, line 108: Is a partial match ok in this context? For example, it would match with deselect-one as well.

CHANGELOG.MD
```diff
@@ -1,3 +1,14 @@
+## July 27, 2026
+
+- **Feature** Re-style Formio components; ensure BC Gov components work properly [🎟️ DEP-276](https://citz-gdx.atlassian.net/browse/DEP-276)
+  - Ensured all Formio imports come from the same package, which means all components are brought into the same runtime when using Formio.use(). This fixes an issue where some components were unable to render certain other components as children.
+  - Updated Formio component styles to match BC Gov design system, including button sizes, focus states, and other visual elements. This ensures a consistent look and feel across the application. Also imported some style overrides from the CHEFS frontend.
+  - Created custom overrides to ensure focus states and buttons feel at home within DEP.
+  - Added focusability for buttons that lacked tabindex and aria roles, improving accessibility for keyboard and screen reader users.
+  - Removed kludged Bootstrap styles and imported Bootstrap SCSS directly into the Formio SCSS file, allowing for better control over styling and ensuring Bootstrap does not overwrite outer components.
+  - Removed buggy Typescript declarations for Formio.Utils; import directly from Formio instead
```
jareth-whitney, line 9: So odd that we were defining these types ourselves before, but way to reduce overhead.

jareth-whitney (approved):
Thanks for sifting through the styles and making them coherent for these form components. Having components that follow the BC design system, and a good variety of form components, will greatly enhance the survey builder. Nothing blocking, just take a quick look at the following:

- event.currentTarget value guaranteed in all scenarios?
- tabToFocus.focus() optimization
- getWizardPageTabs() -> could it return a value that can't be iterated?
- Partial match for data type, maybe ok, worth a check

web/src/components/Form/formio.scss
NatSquared: the options are select one and select multi, so this should be fine

web/src/components/Form/FormBuilder.tsx
NatSquared: this cleanup function targets the components the same way we do when adding listeners, so there should be a 100% match rate when cleaning up

web/src/components/Form/FormBuilder.tsx
NatSquared: Even detached nodes should have a currentTarget AFAIK

